Issue #391, opened 2026-05-05
Redesign editor UI around mode-based workflows and contextual panels


fernandotonon:
## Problem Statement

QtMeshEditor has grown a dense, always-on editor UI. The current layout exposes many controls at the same time: object transforms, edit tools, viewport toggles, mesh info, normals, animation tools, material presets, validation, undo history, scene hierarchy, and utility docks all compete for attention.

This makes the editor feel bloated even when the user is doing a focused task such as positioning an object, editing mesh topology, previewing an animation, or validating an import. The impact is visible in several areas:

- The top toolbar mixes core editing commands with specialized actions such as material editing, animation merging, and viewport overlays.
- View toggles such as grid, normals, mesh info, and view cube are always present even though they are secondary display options.
- The right inspector stacks too many sections in one vertical flow, making users scan past unrelated controls.
- Edit mode, scene hierarchy, transforms, animation controls, material presets, validation, LOD controls, and undo history can all appear in the same panel.
- The left/top toolbars expose many icons without enough workflow separation.
- The dark theme and green accent are usable, but the spacing, grouping, and control density make the product feel less modern than comparable professional tools.

The result is higher cognitive load, slower discovery of task-specific tools, and a less professional first impression for a 3D editor.

Reference direction: the UI should move toward the cleaner mode-based layout shown in the provided mock/reference image, while keeping QtMeshEditor's dark theme and green accent as a recognizable identity.

## Proposed Solution

Redesign the editor around progressive disclosure and mode-based workflows, similar in spirit to Blender, Unreal Editor, Godot, and Figma.

Introduce a visible mode switcher with these modes:

- Object Mode
- Edit Mode
- Animation Mode
- Material Mode
- Validation / Scan Mode

Each mode should control:

- Which toolbar actions are visible.
- Which right-inspector sections are shown by default.
- What appears in the bottom context panel.
- Which viewport overlays are enabled by default.

The redesign should preserve existing functionality and shortcuts, but move secondary controls into contextual panels, menus, or viewport overlay popovers.

High-level layout changes:

- Keep the top menu simple and stable.
- Add a clear mode bar below the menu.
- Convert the left toolbar into a contextual tool rail driven by the active mode.
- Make the right panel tabbed and collapsible instead of one long mixed inspector.
- Use a bottom panel for contextual information such as mesh stats, animation timeline, validation results, logs, or scan reports.
[...]

PR #433, opened 2026-05-08
Refine editor mode bar metadata

fernandotonon:
## Summary
- Moves mode-bar button metadata into `EditorModeController` so QML renders a controller-owned list instead of duplicating mode labels/tooltips.
- Adds mode tooltip metadata as an invokable controller API for QML and tests.
- Adds focused tests for mode-bar metadata and main-window status indicator updates when switching modes.

## UX changes & user-visible removals
- The Properties panel no longer shows the green "Edit Mode" strip. With it goes:
  - the inline **Edit / Exit** button (still reachable via `Tab` and the Mode Bar's Edit toggle),
  - the **V / T / S** counts (vertices / triangles / submeshes) that lit up while in Edit mode.
- Mode tooltips on the Mode Bar were also dropped because they were intercepting clicks on the buttons. Tooltip strings are still reachable through `EditorModeController::modeTooltipFor(int)` for accessibility / future consumers.
- The current editor mode now appears as a permanent label on the right-hand side of the main status bar (driven by `updateEditModeIndicator()`), keeping it always visible without claiming Mode-Bar real estate.

## Validation
- `qmllint qml/ModeBar.qml`
- `cmake --build build_local --target UnitTests -j"$(nproc)"`
- `env QT_QPA_PLATFORM=offscreen build_local/bin/UnitTests --gtest_filter=EditorModeControllerTest.*:MainWindowTest.ModeBarLoadsAndModeChangeUpdatesStatusIndicator`
- `cmake --build build_local --target QtMeshEditor -j"$(nproc)"`

## Tracker
- #391, Slice 2

## Summary by CodeRabbit

* **New Features**
  * Mode selection now exposes descriptive tooltips for each editing mode.

* **Improvements**
  * Mode bar is sourced dynamically and its toolbar sizing/docking updated for improved layout.
  * Editor mode shown via a permanent status label (no longer appended to rolling status messages).

* **Bug Fixes / UI Changes**
  * Removed the Edit/Exit toggle and edit-mode indicator from the properties panel.

* **Tests**
  * Added tests validating available modes and mode-bar initialization.

### DIFF
--- a/qml/PropertiesPanel.qml
+++ b/qml/PropertiesPanel.qml
@@ -97,53 +97,6 @@ Rectangle {
                 }
             }
 
-            // ---- Edit Mode Indicator ----
-            Rectangle {
-                width: parent.width
-                height: 32
-                color: EditModeController.editModeActive
-                    ? "#3d6b3d" : PropertiesPanelController.headerColor
-                visible: true
-
-                RowLayout {
-                    anchors.fill: parent
-                    anchors.leftMargin: 8
-                    anchors.rightMargin: 8
-                    spacing: 8
-
-                    Text {
-                        text: EditModeController.modeLabel
-                        color: PropertiesPanelController.textColor
-                        font.bold: true
-                        font.pixelSize: 12
-                        Layout.fillWidth: true
-                    }
-
-                    Text {
-                        text: EditModeController.editModeActive
-                            ? "V:" + EditModeController.vertexCount +
-                              " T:" + EditModeController.triangleCount +
-                              " S:" + EditModeController.subMeshCount
-                            : ""
-                        color: PropertiesPanelController.textColor
-                        font.pixelSize: 10
-                        opacity: 0.7
-                        visible: EditModeController.editModeActive
-                    }
-
-                    Button {
-                        text: EditModeController.editModeActive ? "Exit" : "Edit"
-                        enabled: EditModeController.editModeActive || EditModeController.canEnterEditMode
-                        implicitWidth: 48
-                        implicitHeight: 24
-                        font.pixelSize: 10
-                        ToolTip.text: "Toggle Edit Mode (Tab)"
-                        ToolTip.visible: hovered
-                        onClicked: EditorModeController.toggleObjectEditMode()
-                    }
-                }
-            }
-
             // ---- Selection Target Indicator ----
             Rectangle {
                 width: parent.width
--- a/src/EditorModeController.cpp
+++ b/src/EditorModeController.cpp
@@ -4,6 +4,7 @@
 #include "SentryReporter.h"
 
 #include <QJSEngine>
+#include <QVariantMap>
 
 EditorModeController* EditorModeController::m_pSingleton = nullptr;
 
@@ -83,6 +84,40 @@ QString EditorModeController::modeNameFor(int mode) const
     case ValidationMode: return QStringLiteral("Validation");
     }
     return QStringLiteral("Object");
+}
+
+QString EditorModeController::modeTooltipFor(int mode) const
+{
+    switch (static_cast<Mode>(mode)) {
+    case ObjectMode: return QStringLiteral("Object placement and scene organization");
+    case EditMode: return QStringLiteral("Edit mesh components");
+    case AnimationMode: return QStringLiteral("Animation playback and clip tools");
+    case MaterialMode: return QStringLiteral("Material assignment and editing");
+    case ValidationMode: return QStringLiteral("Mesh validation and scan tools");
+    }
+    return QStringLiteral("Object placement and scene organization");
+}
+
+QVariantList EditorModeController::availableModes() const
+{
+    QVariantList modes;
+    const QList<Mode> orderedModes = {
+        ObjectMode,
+        EditMode,
+        AnimationMode,
+        MaterialMode,
+        ValidationMode
+    };
+
+    for (Mode mode : orderedModes) {
+        QVariantMap entry;
+        entry.insert(QStringLiteral("mode"), static_cast<int>(mode));
+        entry.insert(QStringLiteral("label"), modeNameFor(static_cast<int>(mode)));
+        entry.insert(QStringLiteral("tip"), modeTooltipFor(static_cast<int>(mode)));
+        modes.push_back(entry);
+    }
+
+    return modes;
 }
 
 bool EditorModeController::editModeAvailable() const
--- a/src/EditorModeController.cpp
+++ b/src/EditorModeController.cpp
@@ -109,11 +109,15 @@ QVariantList EditorModeController::availableModes() const
         ValidationMode
     };
 
+    // The map intentionally exposes only `mode` and `label`. ModeBar.qml
+    // dropped tooltips in PR #433 to stop them from intercepting clicks on
+    // top-bar buttons, so a `tip` entry would be dead metadata. Tooltip text
+    // is still reachable through `Q_INVOKABLE modeTooltipFor(int)` for
+    // accessibility hooks and any future consumer.
     for (Mode mode : orderedModes) {
         QVariantMap entry;
         entry.insert(QStringLiteral("mode"), static_cast<int>(mode));
         entry.insert(QStringLiteral("label"), modeNameFor(static_cast<int>(mode)));
-        entry.insert(QStringLiteral("tip"), modeTooltipFor(static_cast<int>(mode)));
         modes.push_back(entry);
     }
 
--- a/src/EditorModeController.h
+++ b/src/EditorModeController.h
@@ -18,6 +18,12 @@ class EditorModeController : public QObject
     Q_PROPERTY(QString modeName READ modeName NOTIFY modeChanged)
     Q_PROPERTY(QString statusText READ statusText NOTIFY statusTextChanged)
     Q_PROPERTY(bool editModeAvailable READ editModeAvailable NOTIFY editModeAvailabilityChanged)
+    // `availableModes` is `CONSTANT` because the set of editor modes is fixed
+    // at compile time. If a future plugin/feature flag ever needs to gate a
+    // mode at runtime (e.g. hiding ValidationMode when the scan engine is
+    // disabled), drop CONSTANT, add an `availableModesChanged()` signal, and
+    // emit it whenever the list changes — otherwise QML bindings on this
+    // property will not refresh.
     Q_PROPERTY(QVariantList availableModes READ availableModes CONSTANT)
 
 public:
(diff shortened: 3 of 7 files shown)


fernandotonon:
Updated the PR after local verification: the mode bar QML widget now has an explicit minimum width and is followed by a toolbar break, so it renders as its own row above the regular top tool toolbar instead of being collapsed into the toolbar area.

Additional validation:
- `qmllint qml/ModeBar.qml`
- `cmake --build build_local --target UnitTests -j"$(nproc)"`
- `env QT_QPA_PLATFORM=offscreen build_local/bin/UnitTests --gtest_filter=MainWindowTest.ModeBarLoadsAndSelectionUpdatesStatusIndicator:EditorModeControllerTest.*`
- `cmake --build build_local --target QtMeshEditor -j"$(nproc)"`

src/mainwindow.cpp
```diff
@@ -1846,6 +1848,11 @@ bool MainWindow::frameEnded(const Ogre::FrameEvent &evt)
     else
         statusMessage += "No selection";
 
+    // Editor mode status (e.g. "Object mode", "Edit mode (Vertex)")
+    const QString modeStatus = EditorModeController::instance()->statusText();
```
fernandotonon: Duplicate mode text in the status bar.

`updateEditModeIndicator()` already drives the **permanent** `m_editModeLabel` from `EditorModeController::statusText()`, and now `frameEnded()` appends the same `statusText()` to the **transient** message via `" | %1"`. Users will see the mode twice — once on the right-hand permanent label and again at the end of the rolling status line.

Suggest picking one source of truth:

- keep this append and remove `m_editModeLabel`'s text update (or the label entirely), or
- keep `m_editModeLabel` and drop this append.

If the duplication is intentional (e.g. permanent label is sometimes hidden under narrow layouts), worth a code comment so future readers don't "fix" it.

src/mainwindow.cpp
```diff
@@ -1566,11 +1566,13 @@ void MainWindow::createModeSurfaces()
         m_modeBar->setResizeMode(QQuickWidget::SizeRootObjectToView);
         m_modeBar->setMinimumHeight(38);
         m_modeBar->setMaximumHeight(38);
-        m_modeBar->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed);
+        m_modeBar->setMinimumWidth(560);
```
fernandotonon: `setMinimumWidth(560)` is a hard pixel minimum. On laptops at 1366×768, side-by-side with the docks, this can squeeze the rest of the toolbar row.

Consider:

- scaling by `logicalDpiX()` / `devicePixelRatio()` so HiDPI users don't get an oversized bar, and
- letting the QML root collapse mode buttons (icons-only) below a threshold instead of forcing the parent widget wider.

Not blocking — just flagging that 560 is a magic number worth documenting.

src/EditorModeController.cpp
```diff
@@ -85,6 +86,40 @@ QString EditorModeController::modeNameFor(int mode) const
...
+        AnimationMode,
+        MaterialMode,
+        ValidationMode
+    };
+
+    for (Mode mode : orderedModes) {
+        QVariantMap entry;
+        entry.insert(QStringLiteral("mode"), static_cast<int>(mode));
+        entry.insert(QStringLiteral("label"), modeNameFor(static_cast<int>(mode)));
+        entry.insert(QStringLiteral("tip"), modeTooltipFor(static_cast<int>(mode)));
```
fernandotonon: `tip` is now dead metadata for the only consumer.

`ModeBar.qml` removed the `ToolTip.text: modelData.tip` binding in this PR, so this string is built every time `availableModes` is read but never displayed. Two options:

1. Restore tooltips in `ModeBar.qml` with a delay (`ToolTip.delay: 600`) and `HoverHandler` so they don't intercept clicks (the original reason for removing them). That keeps this metadata earning its keep.
2. Drop the `tip` field from the map and keep `modeTooltipFor(int)` `Q_INVOKABLE` for future callers / accessibility.

Either is fine; today both paths exist with no consumer.

src/EditorModeController.h
```diff
@@ -17,6 +18,7 @@ class EditorModeController : public QObject
     Q_PROPERTY(QString modeName READ modeName NOTIFY modeChanged)
     Q_PROPERTY(QString statusText READ statusText NOTIFY statusTextChanged)
     Q_PROPERTY(bool editModeAvailable READ editModeAvailable NOTIFY editModeAvailabilityChanged)
+    Q_PROPERTY(QVariantList availableModes READ availableModes CONSTANT)
```
fernandotonon, line 27: `CONSTANT` is correct only if the available-modes list is fixed for the lifetime of the controller.

If a future plugin/feature flag ever gates a mode (Validation hidden when the scan engine is disabled, MCP-only modes, etc.), QML bindings on `availableModes` won't refresh because there's no `NOTIFY`. Worth a short comment here stating the contract ("modes are compile-time fixed; add a signal if this ever becomes dynamic") so the next person doesn't get bitten.

src/mainwindow_test.cpp
```diff
@@ -202,6 +203,32 @@ TEST_F(MainWindowTest, RebuildAllOgreViewportsDoesNotCrash)
     EXPECT_NO_THROW(ow->fsaaSamples());
 }
 
+TEST_F(MainWindowTest, ModeBarLoadsAndSelectionUpdatesStatusIndicator)
```
fernandotonon: Test name doesn't match what it asserts.

`ModeBarLoadsAndSelectionUpdatesStatusIndicator` reads as if it exercises **selection** changes updating the status indicator, but the body only switches modes (`AnimationMode` → `MaterialMode`) and verifies `m_editModeLabel`. Suggest renaming to e.g. `ModeBarLoadsAndModeChangeUpdatesStatusIndicator` — would have helped me when reading the diff for review.

qml/PropertiesPanel.qml
```diff
@@ -97,53 +97,6 @@ Rectangle {
...
-                        implicitHeight: 24
-                        font.pixelSize: 10
-                        ToolTip.text: "Toggle Edit Mode (Tab)"
-                        ToolTip.visible: hovered
-                        onClicked: EditorModeController.toggleObjectEditMode()
-                    }
-                }
-            }
-
             // ---- Selection Target Indicator ----
```
fernandotonon, line 100: Heads-up that this deletion drops in-panel access to:

- the V/T/S counts (vertices / triangles / submeshes) that lived in the green strip when edit mode was active, and
- the inline **Edit / Exit** button.

Users can still toggle via `Tab`, the mode bar, or the soon-to-land permanent status label, but if anyone counted on the V/T/S readout, it's gone after this PR. Worth a line in the PR description / CHANGELOG so the removal is intentional and discoverable.

fernandotonon (commented):
Nice consolidation — single source of truth for mode metadata is the right move. A few inline notes on UX duplication, dead metadata, layout, test naming, and the `CONSTANT` property choice. None are blockers; happy to approve once the duplicate-status question is resolved.

src/mainwindow.cpp
fernandotonon: Done in 78a38b3 — dropped the duplicated append from `frameEnded()`. The permanent `m_editModeLabel` widget remains the single source of truth for mode text; added a comment in `frameEnded()` so the next reader sees why mode isn't appended there.

src/mainwindow.cpp
fernandotonon: Addressed in 78a38b3. The width is now derived from `QFontMetrics::horizontalAdvance("Object Edit Animation Material Validation Edit")` plus per-button chrome, with 560 kept only as a documented floor (`kModeBarMinWidthFloor`) for unusually narrow fonts. HiDPI / accessibility-scaled fonts now scale up automatically without needing a magic pixel value.

src/EditorModeController.cpp
fernandotonon: Went with option 2 in 78a38b3 — `tip` is no longer included in the `availableModes` map, so we don't pay to build it on every read. `Q_INVOKABLE QString modeTooltipFor(int)` stays as the public API for accessibility tooling and any future consumer that wants the descriptive text. Added a comment at the loop site explaining the contract.

src/EditorModeController.h
fernandotonon: Added a contract comment above the `Q_PROPERTY` in 78a38b3 stating that modes are compile-time fixed and that anyone making them dynamic must drop `CONSTANT`, add an `availableModesChanged()` signal, and emit on changes.

src/mainwindow_test.cpp
fernandotonon: Renamed in 78a38b3 to `ModeBarLoadsAndModeChangeUpdatesStatusIndicator` to match what the body actually exercises (mode switches → status indicator), and updated the PR validation block to reference the new name.

qml/PropertiesPanel.qml
fernandotonon: Documented in the PR body under a new **UX changes & user-visible removals** section (updated alongside 78a38b3). Calls out: the inline Edit/Exit button is gone (still reachable via Tab and the ModeBar Edit toggle), V/T/S counts no longer surface in the panel, and the current mode now lives as a permanent label on the status bar.

fernandotonon:
Addressed all six review threads in 78a38b3:

- **Duplicate mode text in status bar** → dropped the rolling-message append in `frameEnded()`. The permanent `m_editModeLabel` widget remains the single source of truth (added an explanatory comment).
- **Magic 560 minimum width** → derived from `QFontMetrics` so HiDPI / accessibility-scaled fonts scale up automatically; 560 is now a documented `kModeBarMinWidthFloor` only.
- **Dead `tip` metadata** → removed from `availableModes` map. `Q_INVOKABLE modeTooltipFor(int)` stays as the public API for accessibility / future consumers.
- **`CONSTANT` contract** → added a comment above the `Q_PROPERTY` documenting that anyone making modes dynamic must swap to NOTIFY.
- **Test rename** → `ModeBarLoadsAndSelectionUpdatesStatusIndicator` → `ModeBarLoadsAndModeChangeUpdatesStatusIndicator` (validation block in PR body updated too).
- **V/T/S + Edit/Exit removal** → called out under a new **UX changes & user-visible removals** section in the PR body.

Also picked up the CodeRabbit nitpicks:
- Extended `AvailableModesExposeModeBarMetadata` to assert `label` for every mode (and explicitly check that `tip` is no longer exposed); added `ModeTooltipsCoverAllModes` for the accessibility-only path.
- Added `EditorModeController::kill()` to `MainWindowTest::TearDown` so subsequent test cases see a fresh ObjectMode controller.

PR #437, opened 2026-05-08
Add contextual tool rail actions

fernandotonon:
## Summary
- Add mode-scoped left rail actions for animation, material, and validation workflows.
- Reuse existing actions/controllers for Dope Sheet, Curve Editor, Merge Animations, Material Editor, and MeshValidator.
- Keep shared menu actions reachable while hiding their top-toolbar widgets by mode.
- Add coverage for rail mode switching, shared action reachability, and bottom-tool button behavior.

## Validation
- `cmake --build build_local --target UnitTests -j"$(nproc)"`
- `env QT_QPA_PLATFORM=offscreen build_local/bin/UnitTests --gtest_filter=MainWindowTest.ContextualToolRail*:MainWindowTest.ModeBarLoadsAndModeChangeUpdatesStatusIndicator`
- `cmake --build build_local --target QtMeshEditor -j"$(nproc)"`

## Release Notes

* **New Features**
  * Added quick-access toolbar buttons for Dope Sheet, Curve Editor, Merge Animations, Material Editor, and Mesh Validation tools.

* **Improvements**
  * Toolbar now intelligently displays and hides relevant editing tools based on your current mode.
  * Toolbar state automatically updates when your selections change.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1808,10 +1808,14 @@ void MainWindow::updateToolRailForMode()
     const bool materialMode = mode == EditorModeController::MaterialMode;
     const bool validationMode = mode == EditorModeController::ValidationMode;
 
-    auto setSharedToolbarActionVisible = [this](QAction* action, bool visible) {
+    // Keeps the QAction reachable in menus (so users can always find Material
+    // Editor / Merge Animations under their menu) while showing or hiding only
+    // the toolbar-button representation per the active mode. The QAction's
+    // own visibility is intentionally left unchanged.
+    auto setSharedToolbarButtonVisible = [this](QAction* action, bool showButton) {
         action->setVisible(true);
         if (QWidget* toolbarButton = ui->toolToolbar->widgetForAction(action))
-            toolbarButton->setVisible(visible);
+            toolbarButton->setVisible(showButton);
     };
 
     for (QAction* action : ui->objectsToolbar->actions()) {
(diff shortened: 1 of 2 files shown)


fernandotonon:
Addressed both CodeRabbit findings in 119d199:

- **Misleading `setSharedToolbarActionVisible`** (`src/mainwindow.cpp` ~L1811) — renamed to `setSharedToolbarButtonVisible` with a comment documenting the contract. The QAction stays visible in menus on purpose so users can always reach Material Editor / Merge Animations from the menu bar, while only the toolbar-button widget is gated by the active mode. The parameter name (`showButton`) now reflects that.
- **Redundant `updateToolRailForMode()` in `updateMergeAnimationsButton`** (`src/mainwindow.cpp` ~L2596-2618) — removed all three calls. `SelectionSet::selectionChanged` is wired to `updateMergeAnimationsButton` first and `updateToolRailForMode` immediately after (see the `connect()` pair around L1402-1404), and Qt fires connections in registration order — so the rail is already refreshed for free after this slot completes. Added a comment in `updateMergeAnimationsButton` so the next reader sees why we don't call it here. Existing `UpdateMergeAnimationsButton*` and `ContextualToolRail*` tests still cover the behaviour through the signal path.

Validation:
- `cmake --build build_local --target QtMeshEditor UnitTests -j"$(nproc)"` (clean)
- Local GUI smoke test (app starts, no console errors)

PR #646, opened 2026-05-20
fix(export): glTF/GLB skip ConvertToLeftHanded — fixes VAT mesh artifacts

fernandotonon:
## Summary

The main Assimp exporter path applied `aiProcess_ConvertToLeftHanded`
for every format except DirectX (`.x`). glTF is right-handed like
Ogre, so the conversion flips the X axis **and** reverses triangle
winding on every exported `.gltf` / `.glb`.

For a typical round-trip (re-import the file back into another DCC
tool) this was silent — the importer just sees a mirrored mesh and
doesn't care. It became visible the moment we paired the glTF with a
**separately-computed vertex stream**: the OpenVAT bake.

`qtmesh vat` writes two artifacts that **must** agree on vertex order:

| Artifact | Source | Handedness (before fix) |
|---|---|---|
| `<anim>_pos.png` | Ogre vertex buffer, RH | RH (✓) |
| `source.gltf` | Assimp export, post-`ConvertToLeftHanded` | LH (✗) |

The two were in different coord spaces, so the Godot/Unity/Unreal
shader (which fetches positions from the texture by vertex index)
displaced every vertex against a mirrored mesh — exactly the
"weird artifacts" the user reported. From the diff of two `source.gltf`
files (pre-fix vs post-fix), every accessor min/max was swapped on
the X axis:

```diff
-        0.33858245611190796   # max.x (pre-fix, LH-flipped)
+        0.24923135340213776   # max.x (post-fix, RH)
-        -0.24923135340213776  # min.x (pre-fix)
+        -0.33858245611190796  # min.x (post-fix)
```

The pose exporter (`exportCurrentPose` at line 3080) already had the
correct skip list (`"x" || "gltf2" || "glb2"`); this PR brings the
main `exporter()` path (line 2804) into alignment.

## Bundled demo re-bake

Also re-runs the website's VAT demo bake + Godot web re-export so
the live demo at `website/public/demo/index.html` reflects the fix
on the next deploy. The OpenVAT sidecar JSON is **byte-identical**
(bounds unchanged — the bake itself was always RH and never wrong);
only `source.gltf` and the web-export `.pck` move.

## Test plan
- [x] `cmake --build build_local --target QtMeshEditor` — clean
- [x] Re-baked `rumba115/Rumba Dancing.fbx` with the fixed binary;
  resulting `source.gltf` has un-flipped accessor mins/maxes
- [x] Godot web re-export uses the corrected assets
- [ ] Smoke the website demo iframe after deploy to confirm no
  more "weird artifacts" on the showcase or perf scenes

## Related
Builds on top of master after #640. Independent of #644 (the
website tabs/MultiMesh PR) — both can land in either order.

## Summary by CodeRabbit

* **New Features**
  * VAT baker and tools can align baked VAT textures with exported mesh vertex order, emit a bind sidecar, and remap Godot meshes at load when available; new option to provide a vertex permutation.

* **Bug Fixes**
  * Improved material import/export handling for diffuse units and corrected glTF/glb handedness/UV flipping; fixed normal-handedness in VAT playback; adjusted demo materials' specular.

* **Chores**
  * Updated demo export preset and demo package file size.

[Review Change Stack](https://app.coderabbit.ai/change-stack/fernandotonon/QtMeshEditor/pull/646?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


fernandotonon:
@coderabbitai all three notes from this pass were already addressed in commit `0b33e62` (the one immediately before this scan). I think the threads re-fired because the line numbers shifted under the changes:

- **VATBaker bijection check** — added at `src/VATBaker.cpp:545-561` (range/uniqueness scan before `packOpenVAT16`).
- **`sourceMeshMatchesBake` gating** — added at `src/CLIPipeline.cpp:5504`; JSON `sourceMesh` and the "vertex order matches the bake" text are now both gated behind it, with explicit warnings on every fallback path.
- **Full-signature matching** — `buildVertexPermutation` now keys on `(position, normal, UV0)` with two-tier quantization (1e-5 for positions, 1e-3 for normals/UVs to absorb Assimp's re-normalisation drift), V-flipped UVs to undo Assimp's implicit `aiProcess_FlipUVs`, and refuses ambiguous buckets instead of FIFO-picking.

Verified end-to-end against Rumba Dancing.fbx: the matcher walks all 5828 verts, finds unique signatures for every one, permutation is identity for this asset (Assimp's JoinIdenticalVertices happens to be a no-op here), and `mesh: ... (vertex order matches the bake)` is correctly advertised.

Issue #658, opened 2026-05-22
PS1: Per-draw GTE matrix and drawing environment at primitive submit


fernandotonon:
Parent epic: #412

## Problem

Captured primitives carry **screen-space** coordinates. Reconstruction uses `prim.matrixId` from **latestMatrixId** / RAM-scanned matrices (`#419`), then `GteInverse::screenToModel`. When inversion fails or radius exceeds threshold, vertices fall back to `psxScreenToWorld` → all tris collapse into a **screen-centered blob**.

Retail games need the **matrix + drawing mode active at the GPU draw**, not an inferred “latest” matrix from a COP2 RAM scan.

## Goal

Tag each captured primitive with the **correct** projection/model context at submit time so `MeshReconstructor` can place vertices in a stable model/world space for at least some commercial static scenes.

## Scope

- At GP0 dispatch (hook path) or RAM ingest, record:
  - GTE matrix snapshot id (or inline `MatrixRecord`) **per primitive**
  - Relevant `DrawModeRecord` / TPAGE / semi-trans / texture window already partially present — ensure consistency at draw time
- Prefer matrix **stack** or “current environment” updated on `0xE1`–`0xE6` drawing commands, not only post-hoc RAM scan on Capture Frame.
- Reduce `psxScreenToWorld` fallback rate on golden-suite scenes (metric: % vertices using successful `screenToModel`).
- Tests: synthetic multi-matrix frame → reconstructed parts land in separated bounds; integration golden scene checklist.

## Acceptance criteria

- [ ] Golden-suite commercial scene: reconstructed mesh **bounds are not a single screen-space slab** (manual: extent in X/Y/Z similar order of magnitude, recognizable silhouette).
- [ ] Status/log reports matrix association stats (e.g. % prims with valid inverse).
- [ ] Document limitations (dual-pass, shared matrix across OT) in `PS1_RIP_DESIGN.md`.

## Depends on

- #419, #422, in-core GP0 capture issue

## Related

- #424 (coordinate / affine UV normalization)

PR #675, opened 2026-05-26
feat(ps1): real GTE math + orthonormal matrix gate + matrix-tag diagnostic (#675)

fernandotonon:
## Summary

Closes the screen-space half of the "real meshes from real PS1 games" roadmap (PR #674 was the model-space half via TMD/HMD RAM scanner). With this PR the screen-space capture path no longer collapses to a flat-XY blob whenever the GTE matrix isn't the identity.

### Why the blob existed

Both `GteInverse::modelToScreen` and `GteInverse::screenToModel` were diagonal-only (`vx / RT[0][0]`, `vy / RT[1][1]`, `vz / RT[2][2]`). That's not the PSX GTE math — RTPS/RTPT does a full `IR = RT * V + TR` 3x3 multiply. The existing roundtrip test in CI only used an identity-on-diagonal matrix, so the bug was silent. Every non-identity rotation got `(huge, huge, huge)` from `screenToModel`, the `kMaxVertexRadius = 64` filter in `MeshReconstructor::vertexFromPsx` rejected it, and `psxScreenToWorld` painted a flat XY blob.

Secondary issue: the matrix RAM scanner accepted ~192 false-positive "matrices" in a single-box scene (per the screenshot the user shared during the #674 conversation) because the validator only checked per-entry magnitude.

Tertiary issue: `MeshReconstructionStats::primsWithMatrixId / primsTotal` lived in the `ps1.rip.matrix.stats` Sentry breadcrumb but never reached the status bar, so users had to read Sentry to triage which subsystem was failing.

### Changes

**GTE math** — `src/PS1/runtime/GteInverse.cpp`
- `modelToScreen` now implements the real psx-spx RTPS: `IR[r] = (sum_c RT[r][c] * V[c]) / 4096 + TR[r]`, then standard projection through `H` / `OFX` / `OFY`.
- `screenToModel` inverts via `IR[2] = sz`, `IR[i<2] = (screen_i - OF_i) * IR[2] / H`, then `V = RT^T * (IR - TR)`. The `RT^T` fast path is valid because PS1 RT matrices are orthonormal rotations in 12.4 fixed point.

**Orthonormal validator** — `src/PS1/runtime/GteCapture.{h,cpp}`
- New `GteCapture::looksOrthonormalRotation(MatrixRecord)` checks `|row|^2 ~= 4096^2`, `dot(row_i, row_j) ~= 0`, `det ~= +4096^3` (5–10% tolerance for cos/sin quantisation noise).

**Tightened scanner** — `src/PS1/runtime/PsxGteRamScanner.cpp`
- `looksLikeMatrixRecord` narrowed `H in [64, 2048]` (was 50..8192), tightened RT entry magnitude (`|RT[r][c]| <= 8192`), delegated rotation validation to the new orthonormal gate.

**UI diagnostic** — `src/PS1/runtime/PS1RipManager.{h,cpp}` + `PS1RipSessionWindow.{h,cpp}`
- `PS1RipManager::meshBuilt` signal gains `primsWithMatrixId, primsTotal`.
- Status bar now reads `GTE inverse N% (matrix tag X/Y)`. Triage at a glance:
  - `0/Y` = matrix association bottleneck (debug #658 per-draw matrix tagging)
  - `Y/Y` with low % = math/matrix-quality bottleneck (deferred to #676 in-core hook)
  - High % = healthy capture
- `ps1.rip.matrix.stats` breadcrumb already carries the same data — Sentry parity confirmed; no breadcrumb change needed.

### Tests (8 new, all 17 tests across `GteInverse`/`GteCapture`/`PsxGteRamScanner`/`PsxPerDrawMatrix` green)

- `GteInverseTest.ModelToScreenRoundTripsRotated90DegY` — 90deg Y rotation round-trips within 2 fixed-point units.
- `GteInverseTest.ModelToScreenRoundTripsArbitrary3DRotation` — 30deg pitch + 45deg yaw + 15deg roll Euler round-trips within 4 units.
- `GteInverseTest.ModelToScreenForwardChangesAllAxesForRotatedMatrix` — regression catch for the old diagonal-only forward.
- `GteCaptureTest.LooksOrthonormalRotation{AcceptsIdentity,AcceptsRealRotations,RejectsScaledRotation,RejectsReflection,RejectsRandomFill}` — validator behaviour pinned across the four boundary cases.
- `PsxGteRamScannerTest.RejectsNonOrthonormalScaledRotation` — the false-positive class from the #674 screenshot.
- `PsxGteRamScannerTest.Accepts45DegYRotation` — real-game rotation still accepted.
- `PsxGteRamScannerTest.RejectsPseudoRandomGarbage` — 16 KB of pseudo-random RAM produces zero matrices.

### Docs

- `PS1_RIP_DESIGN.md` — new "Screen-space inverse projection (#675)" section covering the math, the orthonormal validator, and the status-bar diagnostic reading order. Refreshed the "Capture mesh is a triangle blob" troubleshoot entry with the new diagnostic ladder (GTE inverse % -> matrix tag X/Y -> tmd N).

### Out of scope (deferred to #676)

- Matrix -> primitive association on retail games still uses heuristic per-draw matrix tagging from #658. #675 makes the inverse **correct when association is correct**; ground-truth association needs the forked `mednafen_psx_libretro` with in-core RTPS/RTPT callback (#676). Custom-engine games (Crash, Spyro, FFVII field models, MGS post-Yaroze) will still produce partial recovery at best until #676 lands — they author custom transform stacks that defeat heuristic matrix association.

## Test plan
- [x] `./build_local/bin/UnitTests --gtest_filter="GteInverseTest.*:GteCaptureTest.*:PsxGteRamScannerTest.*:PsxPerDrawMatrixTest.*"` — 17/17 pass
- [x] Full PS1 test bucket (`*Psx*:*Gte*:*Gp0*:*EmuCore*:CaptureBuffer*:GpuCommand*:MeshReconstructor*:PS1Rip*`) — 104/104 pass; `PS1RipManagerTest.StopCancelsPendingStart` is a pre-existing flaky test (passes in isolation, unrelated to this PR).
- [x] Build full `QtMeshEditor` binary clean.
- [ ] CI: full Linux test matrix + SonarCloud.
- [ ] Manual: open a retail disc via `Tools -> Experimental -> PS1 Runtime Ripper`, arm capture, walk a scene, confirm status bar shows `GTE inverse N% (matrix tag X/Y)` and `N` is meaningfully > 0 on titles where the screen-space path was previously stuck at 0%.

## Release Notes

* **Bug Fixes**
  * Fixed screen-space inverse projection calculations; rotations now properly round-trip between model and screen space.

* **New Features**
  * Enhanced diagnostics display the primitive-to-matrix association ratio in mesh reconstruction status information.

[Review Change Stack](https://app.coderabbit.ai/change-stack/fernandotonon/QtMeshEditor/pull/675?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/PS1/PS1_RIP_DESIGN.md
+++ b/src/PS1/PS1_RIP_DESIGN.md
@@ -125,6 +125,51 @@ See epic #412 for phased issues (#413–#431).
 - **GTE RAM supplement** — Capture Frame still merges COP2-scanned matrices; per-draw tagging applies at GP0 dispatch time, not retroactively to RAM-only captures without draw-env context.
 - **Commercial golden scenes** — manual acceptance tracked in [#659](https://github.com/fernandotonon/QtMeshEditor/issues/659) (`src/PS1/golden_captures.md`); #658 reduces blob fallback but does not replace a full matrix stack or FIFO-accurate stream (#662).
 
+## Screen-space inverse projection (#675)
+
+The screen-space capture paths (live FIFO bridge, OT chains, chain roots, linear scan)
+all deliver post-projection GP0 vertex coordinates — to land them in editor world units
+the reconstructor inverts the PSX GTE projection in `GteInverse::screenToModel`. The
+forward / inverse math matters: when it's wrong, every screen-space vertex falls through
+to `psxScreenToWorld` (a pure pixel-to-flat-XY mapping) and the reconstructed mesh is the
+classic flat-plate "blob".
+
+- **The math:** `GteInverse::modelToScreen` implements the psx-spx RTPS formula
+  `IR[r] = (RT[r][:] · V) / 4096 + TR[r]`, `SX_pixel = H * IR[0] / IR[2] + OFX/65536`
+  (similarly for SY), `SZ = IR[2]`. `GteInverse::screenToModel` inverts it via
+  `IR[2] = sz`, `IR[i<2] = (screen_i - OF_i) * IR[2] / H`, then
+  `V = RT^T * (IR - TR)` (the `RT^T` fast path exploits the fact that PS1 RT matrices
+  are orthonormal — `RT^-1 == RT^T` for unit rotations). Pre-#675 both `modelToScreen`
+  and `screenToModel` were diagonal-only (`vx / RT[0][0]`) which silently passed the
+  identity-matrix roundtrip test in CI but rejected every real rotation matrix at the
+  `kMaxVertexRadius` filter in `MeshReconstructor::vertexFromPsx`.
+- **The orthonormal gate (`GteCapture::looksOrthonormalRotation`):** validates that an
+  `MatrixRecord`'s 3×3 RT block satisfies `|row|^2 ≈ 4096^2`, `dot(row_i, row_j) ≈ 0` and
+  `det ≈ +4096^3` (each within 5–10% slack so 12.4 quantisation of `cos`/`sin` tables
+  doesn't false-reject). Used in two places:
+  - `PsxGteRamScanner::looksLikeMatrixRecord` — drops false-positive matrix candidates.
+    Combined with the narrowed `H ∈ [64, 2048]` and tightened RT entry magnitude
+    (`|RT[r][c]| ≤ 8192`), pseudo-random 16 KB RAM blocks produce **zero** accepted
+    matrices in the regression suite (`PsxGteRamScannerTest.RejectsPseudoRandomGarbage`).
+  - Future: `GteInverse::screenToModel` callers can pre-check this if they want to skip
+    the inverse for non-orthonormal matrices and force `psxScreenToWorld` instead — for
+    now the `RT^T` solve is harmless on any input that survives the scanner gate.
+- **Diagnostics (`MeshReconstructionStats`):** `primsTotal`, `primsWithMatrixId`,
+  `gteInverseVertices`, `screenFallbackVertices`. Plumbed through
+  `PS1RipManager::meshBuilt(... primsWithMatrixId, primsTotal ...)` into
+  `PS1RipSessionWindow`'s status bar as
+  `GTE inverse N% (matrix tag X/Y)`. Sentry breadcrumb `ps1.rip.matrix.stats` carries
+  the same `gte_inverse=N%% prims_with_matrix=X/Y` for off-line analysis.
+- **Acceptance tests** (`GteInverseTest`, `GteCaptureTest`, `PsxGteRamScannerTest`):
+  90° Y rotation and arbitrary 3D Euler (`30° + 45° + 15°`) round-trip within 2–4
+  fixed-point units; identity, real rotations accepted by the validator; scaled
+  rotation, reflection (det = `-4096^3`) and pseudo-random garbage rejected.
+- **Out of scope:** matrix→primitive *association* (which RT was active when this prim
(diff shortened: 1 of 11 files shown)


fernandotonon:
Followup fix pushed (7de6865) for the regression you spotted in the screenshot — "GTE inverse 100% (matrix tag 137/137)" but invisible mesh in the viewport.

**Root cause.** PS1 GP0 polygon packets only carry 2D screen-space `XY`. The GTE writes Z into a separate `SZ` FIFO that the CPU drains *before* assembling the GP0 word, so `PsxVertex::z` is identically 0 for every prim ingested from the GP0 hook / live FIFO bridge (verified: nothing in the codebase ever writes `verts[].z`). With the correct psx-spx math in this PR, `sz=0` makes the pinhole inverse degenerate — `IR[0]=IR[1]=0` and every vertex of a given matrix tag maps to the same point `RT^T·(-TR)/4096`. Mesh exists in the scene tree, has 489 verts / 163 tris, but zero extent → invisible.

Before the math fix the buggy diagonal-only math produced huge radii that the `kMaxVertexRadius=64` filter rejected, so callers fell through to `psxScreenToWorld` and you got the visible (but inaccurate) blob. The math fix made the radius small enough to pass the filter, exposing the latent depth-loss problem.

**Fix.** Early-return `false` from `screenToModel` when `sz=0`, forcing `vertexFromPsx` to call `psxScreenToWorld` — an ugly-but-visible blob beats a degenerate zero-extent mesh. The math fix stays valuable for the `PsxPerDrawMatrixTest` / `MeshReconstructorCubePipelineTest` paths (which inject forward-projected `sz≠0`) and for the eventual #676 in-core GTE callback when retail GP0 captures will actually have depth.

After this commit the status bar on GP0-only captures correctly reads `GTE inverse 0% (matrix tag X/Y)` — which is the truth: matrix association is fine, depth is gone.

**Tests added** (`GteInverse_test.cpp`):
- `ScreenToModelRefusesZeroDepth` — rejects `sz=0` for several screen XY samples
- `ScreenToModelWithZeroDepthDoesNotCollapseVertices` — confirms outputs stay untouched so the caller's fallback drives the per-vertex divergence

**Docs.** `PS1_RIP_DESIGN.md` gains a "GP0 carries no depth" bullet in the screen-space inverse section, an updated diagnostic reading order ("0% on GP0-only captures is expected, see step 3"), and a new troubleshooting entry ("Captured scene tree has nodes but viewport is empty") that pins down exactly this failure mode for future me.

Targeted tests on `feat/ps1-screen-space-inverse-fix-675` after this commit:
- `GteInverseTest` 8/8, `GteCaptureTest` 6/6
- `MeshReconstructorTest` 9/9 (incl. `GtePipelineCubeRoundTripsToUnitCubeMesh`, `GteModelProjectionFeedsReconstruction`)
- `MeshReconstructorGoldenTest` 5/5 (`ScreenCubeReconstructionHasVolume`)
- `PsxGteRamScannerTest` 5/5, `PsxTmdRamScannerTest` 9/9, `PsxHmdRamScannerTest` 3/3
- `Gp0CapturePathsTest` 8/8, `Gp0HookDispatchTest` 1/1
- `CaptureBufferTest` 1/1, `PsxCaptureFiltersTest` 2/2
- `PS1RipManagerTest` 7/7, `PS1RipLegalityDialogTest` 1/1

68 tests, all pass.

PR #718, opened 2026-06-13
fix(macos): keychain prompts, session persistence, texture apply, mesh-tex generation

fernandotonon:
Fixes a cluster of macOS / Material-Editor issues found while testing the AI texture features.

## What was broken & fixed

1. **Repeated macOS keychain prompts at startup (4×).** `CloudCredentialStore` stored the cloud session in the OS secret store (Keychain/Credential Manager/libsecret); every macOS keychain read raises a prompt and the account control reads the session several times at launch. → Session now stored in **QSettings** (per-user prefs, no prompt). Removed the Security.framework/wincred/libsecret code, the `CloudCredentialStore_linux.*` files, and their CMake/link wiring. Kept an in-process cache.

2. **Cloud session not persisting** (logged out each launch). The sign-in handler removed the `cloudToken`/`expiresAt`/`email` QSettings keys right after `saveSession()` wrote them (leftover keychain-era cleanup). → Removes dropped.

3. **Applying a texture didn't show on the model** (only the preview updated):
   - `RTShaderHelper::wirePbrSlotsForFFP` now wires an unnamed/numeric diffuse TUS (e.g. an Assimp-imported `"0"` slot) as the diffuse fallback when there's no named `albedo`/`diffuse_map` slot — previously only named slots got the FFP colour-op, so the texture loaded but never rendered.
   - `MaterialEditorQML::ensureTextureInMaterialGroup` loads the chosen texture into the **material's** resource group before binding, so the RTSS-rendered mesh can resolve it (was resolving against the wrong group → yellow placeholder while the disk-reading preview looked fine).
   - `updateMaterialText` re-applies via `SubEntity::setMaterial(ptr)` instead of `setMaterialName(sameName)` (stale technique pointer).

4. **Plain AI (txt2img) texture not applying:** `onSDGenerationCompleted` now routes the result through `setTextureName()` (group resolution + RTSS rebind) instead of a raw `texUnit->setTextureName()`.

5. **"Use selected mesh (depth-conditioned)" never started:** QML called `MaterialEditorQML.hasSelectedMesh()`, but it's a `Q_PROPERTY` → the call threw "not a function" and silently aborted the handler. Now read as a property. Missing-ControlNet emits a non-fatal `sdGenerationNotice` (was `sdGenerationError`, which reset the run) so it proceeds in prompt-only mode. The checkbox no longer unchecks itself when generation starts (auto-uncheck keys off selection loss, not the disabled state).

6. **Logged-out account icon blank in deployed builds:** the avatar used an SVG `QIcon` that renders empty without the Qt SVG icon-engine plugin (showed locally, vanished in the installed app). The glyph is now drawn with `QPainter` for both states — no plugin dependency.

7. **Docs:** Homebrew install now includes `brew trust fernandotonon/qtmesheditor` (Homebrew now requires trusting third-party taps before loading their casks — this is the new expected workflow, not a cask bug).

## Testing
- `CloudCredentialStore_test.cpp` updated for the QSettings backend (round-trip, cache, no-op migration).
- Verified interactively on macOS: no keychain prompts, session persists across restarts, manual + plain-AI + mesh-conditioned texture generation all apply to the model, logged-out avatar renders.

## Summary by CodeRabbit

* **New Features**
  * Added non-fatal status notices for Stable Diffusion generation events.

* **Bug Fixes**
  * Fixed mesh-conditioned texture generation and checkbox behavior during generation.
  * Improved material texture resolution and rendering; better handling for materials lacking explicit diffuse slots.
  * Updated account/avatar rendering to show a generic profile glyph when appropriate.

* **Documentation**
  * Added macOS Homebrew trust step to installation docs.

* **Chores**
  * Bumped project release to 3.5.3; updated packaging and publishing workflow and reduced reliance on OS secret-store integration.

* **Tests**
  * Updated tests and added a cache-reset hook for session handling.


fernandotonon:
Addressed all review feedback in 175fbfa:

- **Preserve existing secret-store sessions during migration** (Codex P2): `migrateLegacySettingsIfNeeded()` now does a one-time read of the old macOS Keychain / Windows Credential Manager + the legacy `cloud_session.dat` fallback and copies it into QSettings, so users upgrading from a pre-QSettings build stay signed in. Re-added the macOS `Security` / Windows `advapi32` link for that read only; libsecret stays dropped.
- **Treat tokenless settings as an empty session** (CodeRabbit): `readFromSettings()` returns `{}` immediately when no token is present — no more partial sessions / stale `email`.
- **Propagate `QSettings::sync()` failures from `saveSession()`** (CodeRabbit): `writeToSettings()` now returns `status() == NoError` and `saveSession()` propagates it (and doesn't prime the cache on failure).
- **Use `isAlbedoSlotName()` for diffuse detection/wiring** (CodeRabbit): both spots in `wirePbrSlotsForFFP` now use the helper (covers `Diffuse`/`BaseColor`), so the unnamed-diffuse fallback can't mis-wire an aliased pass.
- **`file.import` breadcrumb in `ensureTextureInMaterialGroup`** (CodeRabbit): added on success and failure.
- **Remove stale `libsecret-1-0` Debian dependency** (CodeRabbit): dropped from `DEBIAN-control.in`.

Added tests for the legacy-fallback-file migration and the tokenless-settings case. All changed files compile; the only local link failure is the pre-existing `qHash(TextureDecoder::MaterialKey)` in the PS1 runtime (unrelated, present on master, macOS-only — CI runs tests on Linux).

fernandotonon:
Addressed in 3705dc4: added the `-framework Security` / `advapi32` link to the **UnitTests** target too (it compiles `CloudCredentialStore.cpp` and the test exercises `migrateLegacySettingsIfNeeded`). Also fixed a latent test-build issue: `src/PS1/runtime/*_test.cpp` were globbed unconditionally while their impl is gated on `ENABLE_PS1_RIP`, so a tests build with the flag off failed to link (`qHash(TextureDecoder::MaterialKey)`); those tests are now gated the same way. UnitTests links cleanly locally now.

Issue #403, opened 2026-05-06
AI: Mesh-aware texture generation (depth-conditioned sd.cpp)


fernandotonon:
Parent epic: #397

## Goal

Generate textures that **fit the mesh** by conditioning sd.cpp on rendered depth + normal maps from the current viewport, then bake the result back to UVs. The single biggest user-visible AI win for the texturing workflow.

## Why

The existing `SDManager` produces flat 2D textures that often don't fit the mesh's UV layout or geometry. Depth/normal-conditioned generation (ControlNet-style) produces textures where lighting and details follow the mesh shape, which is what indie devs actually want.

## Scope

- Render mesh depth + normal maps from the active viewport using Ogre RTT.
- Feed those into `sd.cpp` as ControlNet conditioning images (sd.cpp supports ControlNet).
- Bake the generated texture to the mesh's UVs and assign it to the active material's diffuse slot.
- GUI: a new "Generate from mesh" action in Material Mode that opens a prompt dialog and shows progress through `SDManager`'s existing progress signal.
- CLI: `qtmesh material model.fbx --generate-texture "rusty bronze armor" -o out.fbx`.
- MCP: `generate_mesh_texture` tool with `prompt`, `model`, `controlnet_strength` params.
- Pair with the PBR map synthesis task to optionally produce normal/roughness/height alongside the diffuse.

## Acceptance Criteria

- [x] Generated textures show silhouette and lighting that match the mesh shape.
- [x] Works on at least one humanoid and one prop test mesh.
- [x] Progress + cancellation work through the existing `SDManager` worker pattern.
- [x] CLI and MCP variants produce the same output as the GUI for the same params.
- [x] Documented limitation: requires an existing UV layout (suggest running `qtmesh uv --unwrap` first if needed).
- [x] Sentry breadcrumb `ai.assist.mesh_texture` recorded.

## Dependencies

- ControlNet support in sd.cpp (already supported upstream; verify the version we link against).
- Working UV layout on the input mesh (#400).

## Effort

~7 days.

fernandotonon:
## Done ✅

All acceptance criteria are met across these PRs:

| Criterion | Delivered in |
|---|---|
| Silhouette/lighting follows mesh shape (depth-conditioned ControlNet) | #717 |
| Works on humanoid + prop meshes | #717 (Rumba humanoid + props) |
| Progress + cancellation via the `SDManager` worker pattern | #717 |
[...]

PR #729, opened 2026-06-16
feat(texture): multi-view AI texture bake (front+back projection onto UV0)

fernandotonon:
Generates a diffuse texture by ControlNet-generating images from multiple camera views (front + back to start) and **projection-baking** them onto the mesh's existing UV0 atlas — replacing the prior planar bind that was unreliable on RTSS/PBR materials (#403 known limitation). See `src/MULTIVIEW_TEXTURE_BAKE_DESIGN.md`.

Key idea: the UV layout stays fixed; each view's generated image is re-projected through the **exact camera that produced its depth map** and rasterized into UV0 (this is a bake, not "recompute UVs from the image").

## Slice 1 — reprojection foundation
- `MeshDepthRenderer::renderDepthMapView(entity, size, View)` returns the depth image **plus the exact view/projection matrices + camera dir** used. Built-in front/back/left/right/top/bottom views. `renderDepthMap()` stays as the front-view delegate so #403 is unchanged.
- `MultiViewTextureBaker` (pure-data): per triangle × view, facing-weight cull (`max(0,-n·viewDir)^p`), project to viewport UV, rasterize in UV0 space, sample the view image at the barycentric-interpolated screen UV, accumulate `color×weight`; resolve by weight + seam-dilate (reuses `VertexColorBaker::dilate` + `TexturePaintBuffer`).
- Unit tests: synthetic ±Z quad + solid images assert front paints, back-only is culled, front+back blends by facing, dilation fills margins, empty-input guards. (Math also validated locally via a standalone harness.)

## Slice 2 — front+back orchestration + UV0 bake
- `MultiViewTextureBaker::fromEntity()` extracts world-space triangles + UV0 from a live entity (world matrix on positions, inverse-transpose on normals).
- `MaterialEditorQML::generateMeshTextureMultiView(prompt,w,h,strength,views)`: a state machine that renders each view's depth+matrices, generates one image per view sequentially through the SD worker, and on the final completion bakes the accumulated images onto UV0 and applies the atlas to the entity's diffuse. `onSDGenerationError` aborts cleanly. Defaults to `{front, back}`. `ENABLE_STABLE_DIFFUSION`-guarded.
- QML: "Bake front + back (projects onto UV map)" sub-toggle under the existing "use selected mesh" checkbox.

## Incidental fixes found while building this
- **Keychain prompt regression**: `migrateLegacySettingsIfNeeded()` re-probed the OS secret store on every signed-out launch (SecItemCopyMatching → macOS prompt). Now gated behind a persisted `Cloud/legacyMigrationDone` flag → probes at most once, ever. Regression test added.
- **Scene tree**: hide the persistent offscreen `QtMeshDepthCameraNode` from the Inspector (added to `Manager::isForbiddenNodeName`).

## Testing
- App + UnitTests build clean (macOS); manually verified front+back bake produces a correct texture on the model.
- Pure-data baker covered by unit tests; SD-driven orchestration is `#ifdef`-guarded (no model in CI), matching #403.

## Release Notes

* **New Features**
  * Added multi-view AI texture baking that projects front and back views onto the mesh UV0 atlas for improved coverage and consistency.
  * Added **“Bake front + back (projects onto UV map)”** option to the texture generation UI (shown when mesh conditioning is available).

* **Bug Fixes**
  * Reduced repeated legacy cloud credential prompts by recording a one-time migration completion flag and avoiding repeated legacy probing.

* **Other**
  * Improved filtering of a depth-related scene node during processing.

### DIFF
--- a/qml/TexturePropertiesPanel.qml
+++ b/qml/TexturePropertiesPanel.qml
@@ -20,7 +20,11 @@ GroupBox {
         var useMesh = useMeshCheck.checked && MaterialEditorQML.hasSelectedMesh
         var w = sdWidthSpin.value
         var h = sdHeightSpin.value
-        if (useMesh) {
+        if (useMesh && multiViewCheck.checked) {
+            // Front+back projection bake onto the mesh's UV0 (slice 2).
+            MaterialEditorQML.generateMeshTextureMultiView(
+                sdPromptField.text, w, h, meshStrengthSlider.value, ["front", "back"])
+        } else if (useMesh) {
             MaterialEditorQML.generateMeshTextureFromPrompt(
                 sdPromptField.text, w, h, meshStrengthSlider.value)
         } else {
@@ -915,6 +919,22 @@ GroupBox {
                                     useMeshCheck.checked = false
                             }
                         }
+                    }
+                }
+
+                // Slice 2: multi-view bake. When on (and mesh conditioning is
+                // enabled), generate front + back images and projection-bake
+                // them onto the mesh's UV0 atlas instead of a single planar
+                // view. Only meaningful with "use selected mesh" checked.
+                RowLayout {
+                    Layout.fillWidth: true
+                    spacing: 8
+                    visible: useMeshCheck.checked && useMeshCheck.enabled
+                    ThemedCheckBox {
+                        id: multiViewCheck
+                        text: "Bake front + back (projects onto UV map)"
+                        enabled: !MaterialEditorQML.sdIsGenerating
+                        checked: false
                     }
                 }
 
--- a/src/AppSettingsKeys.h
+++ b/src/AppSettingsKeys.h
@@ -98,6 +98,15 @@ inline const QString& cloudUserSlug()
     return k;
 }
 
+/** @brief Set once the one-time pre-QSettings secret-store migration has been
+ *  attempted, so the OS keychain is never probed again (which would prompt on
+ *  every launch). */
+inline const QString& cloudLegacyMigrationDone()
+{
+    static const QString k(QStringLiteral("Cloud/legacyMigrationDone"));
+    return k;
+}
+
 /** @brief Last platform profile id used for Validation-mode asset folder scan (issue #370). */
 inline const QString& validationPlatformProfileId()
 {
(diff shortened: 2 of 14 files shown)


fernandotonon:
Addressed the review feedback in 13a5f63:

- **Front depth view on the -Z side (Codex P1) / flip built-in view directions (CodeRabbit major):** correct — with `camPos = center - dir*dist`, `front` must be `dir=(0,0,1)` to place the camera on -Z (the front for -Z-facing imports). The previous `(0,0,-1)` put it on +Z and captured the back, regressing the #403 single-view path and swapping front/back in multi-view. Flipped all six views + comments; this also fixes the `camDirection` fed to the baker's facing cull.
- **Return failure when UV0 is unusable (CodeRabbit major):** `fromEntity` now returns an empty triangle list (not just sets errorOut) so the caller aborts instead of baking onto degenerate UVs.
- **Abort if a per-view image can't be loaded (CodeRabbit major):** `onSDGenerationCompleted` now resets the multi-view state and errors immediately on a null `QImage` rather than advancing with empty pixels.
- **Clear multi-view state on Stop (Codex P2):** `onSDGenerationStopped` now resets `m_multiViewBake`.
- **Breadcrumb categories (CodeRabbit):** added `ui.action` on the trigger and `file.export` on the baked-PNG write.
- **Sanitize entity name for filename (CodeRabbit):** baked PNG basename is now `[^A-Za-z0-9._-] → _` sanitized.
- **Test stabilization + markdownlint (CodeRabbit):** clear the legacy `cloud_session.dat` before the first migrate in the one-time-probe test; tagged the doc's ASCII fence as `text` and added the table blank line.

**Skipped:** adding Sentry breadcrumbs inside `CloudCredentialStore` — the store is intentionally dependency-free (QSettings/QFile/Security only); coupling it to `SentryReporter` for a low-value breadcrumb isn't worth it. The user-facing actions that call it already emit breadcrumbs at the UI layer.

App + UnitTests build clean.

fernandotonon:
Addressed the absolute-path breadcrumb finding in abdc81e: the `file.export` breadcrumb now logs the baked-texture basename (`outName`) instead of the absolute `outPath`, so no local username/path leaks into telemetry.

Issue #458, opened 2026-05-09
Epic: UV — UV Map editing


fernandotonon:
## Overview

QtMeshEditor can read and write UV coordinates today (`EditableMesh::setVertexUV`/`getVertexUV` at `src/EditableMesh.h:518`, full UV0 round-trip through `MeshImporterExporter`, the FBX exporter, and the validator), but there is **no way for a user to view or edit the UV layout in the editor**. The only UV-related feature on the roadmap right now is auto-unwrap (#400, xatlas).

This epic plans a proper **UV editor**: a 2D panel where the user can see the UV islands of the active selection, select UV components, transform them (move/rotate/scale, with snapping and pinning), seam/sew along edges, project from view/camera/box/cylinder/sphere, pack islands, and switch between UV channels — fully integrated with the existing Edit Mode, undo system, and CLI/MCP surfaces.

This is a **GUI-led** epic (most value lives in the interactive panel), but each meaningful operation must also be exposed via CLI and MCP to match project conventions.

## Why

- Auto-unwrap (#400) gets users a starting layout, but real workflows need manual cleanup: rotating shells, scaling for texel density, fixing overlaps, marking seams, pinning verts before re-unwrap.
- Mesh-aware texture generation (#403) and PBR map synthesis (#404) both depend on a clean UV layout. Without an editor, users have to round-trip through Blender — defeating the offline single-binary value prop of QtMeshEditor.
- The infrastructure is mostly already there: `EditableMesh` exposes per-vertex UVs, `HalfEdgeMesh` provides adjacency, the `EditModeController` already manages Vertex/Edge/Face component selection, and the QML inspector + Tab-mode toggle are reference patterns. The missing piece is the 2D viewport and the UV-specific selection/transform layer.

## Architecture

A new singleton, **`UVEditorController`** (mirrors `EditModeController` and `PropertiesPanelController`):

- Owns the UV editor's selection state (UV-vertex / UV-edge / UV-face), independent from the 3D component selection but kept in sync when desired.
- Reads/writes through `EditableMesh::setVertexUV`/`getVertexUV`. **Source of truth remains `EditableMesh`.**
- Uses `HalfEdgeMesh` to detect UV islands, seams, and boundary loops.
- Pushes a single `UVEditCommand` per operation onto the existing `UndoManager` (`Ctrl+Z`/`Ctrl+Shift+Z`).
[...]

PR #757, opened 2026-06-23
fix(import): restore textured viewport and paint preview on File→Open

fernandotonon:
## Summary
- Rebind RTSS materials for entities imported via **File → Open** (`importMeshs`), matching the post-import path cloud downloads already used — fixes white/untextured models in the 3D viewport on Linux.
- Extract `textureSearchRootsForImportFile()` so cloud and local imports share the same sidecar texture search roots.
- Fix blank **Texture Paint** 2D preview by loading the paint buffer from embedded FBX bytes and on-disk paths before GPU readback (which is unreliable for imported textures).

## Scope
This PR intentionally **does not** include UV Editor work (epic #458). UV editor will be restarted slice-by-slice in follow-up PRs.

## Test plan
- [ ] File → Open an FBX with external textures (e.g. Mixamo character) — 3D viewport shows diffuse maps, not flat white
- [ ] Enable Texture Paint on that mesh — 2D preview shows the albedo, not a blank/white canvas
- [ ] Cloud project download import still shows textures (regression)
- [ ] Linux build (`QtMeshEditor` target)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ cmake_minimum_required(VERSION 3.24.0)
 cmake_policy(SET CMP0005 NEW)
 cmake_policy(SET CMP0048 NEW) # manages project version
 
-project(QtMeshEditor VERSION 3.9.1 LANGUAGES C CXX)
+project(QtMeshEditor VERSION 3.9.2 LANGUAGES C CXX)
 message(STATUS "Building QtMeshEditor version ${PROJECT_VERSION}")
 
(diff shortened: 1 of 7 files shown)


fernandotonon:
Addressed Codex P2 feedback in b2fd106: texture rebind now uses each entity's `qtme.source_path` via `textureSearchRootsForEntity()`, so multi-file File→Open cannot cross-bind common texture basenames.

fernandotonon:
Merged latest `master` (Xcode 26.3 pin + `MACOS_CACHE_VERSION` bump) to fix the `libz.tbd` / `build-n-cache-ogre-macos` CI failure.

Issue #460, opened 2026-05-09
UV: Slice B — UV component selection (vertex/edge/face)


fernandotonon:
Parent epic: #458
Depends on: Slice A

## Goal

Add UV-vertex / UV-edge / UV-face component selection inside the UV editor panel, independent from but optionally synced with the 3D Edit Mode selection.

## Scope

- Selection state lives on `UVEditorController` — **do not** reuse `SelectionSet`. UV components are inherently distinct from 3D mesh components (an island border duplicates a 3D vertex into multiple UV vertices).
- Shortcuts active only when the UV panel has focus: `1` = UV-vertex, `2` = UV-edge, `3` = UV-face.
- Click, Shift-click (add), Ctrl-click (toggle), drag-box select.
- Visual feedback: highlight selected components, draw selection box during drag.
- Panel toolbar: explicit mode buttons matching the shortcut keys, current-mode indicator.
- Selection sync toggle (off by default): when on, selecting in 3D propagates a UV selection covering the corresponding UV components, and vice versa. When off, the two selections are independent.
- Highlight the island(s) the active 3D selection belongs to even when sync is off (read-only awareness — different from selection sync).

## Acceptance Criteria

- [ ] All three component-selection modes work via shortcuts and toolbar.
- [ ] Multi-select with Shift/Ctrl works correctly across modes.
- [ ] Drag-box select includes only fully-enclosed components in vertex mode and any-touched in face mode (document the rule).
- [ ] Sync toggle works in both directions without infinite signal loops.
- [ ] Unit tests cover click selection, multi-select, and box select.
- [ ] Sentry breadcrumb `ui.action` recorded on mode change.

## Effort

~4 days.

PR #763, opened 2026-06-27
fix: defer UV rebuild on import to prevent UI freeze

fernandotonon:
## Summary

Fixes UI freezes when importing meshes (especially large/skinned FBX assets).

- **UV Editor (#460 follow-up):** `UVEditorController` was rebuilding the full HalfEdge/island cache + `QVariantList` triangle payload on every `entityCreated` / `selectOne` during import (each sub-entity auto-selects). Now skips rebuilds while the UV Editor dock is hidden, debounces (75ms) when open, and indexes vertices→triangles for edit-mode context islands.
- **Startup/import latency:** Lazy-load hidden dock QML (UV Editor, Asset Browser, AI Chat, Dope Sheet, Curve Editor), defer Asset Browser directory scan + `LLMManager` init, and stop re-applying custom palette via `setPalette()` after QML widgets exist.

## Test plan

- [x] `UnitTests --gtest_filter="UVEditorControllerTest.*"` — 16/16 pass
- [ ] Import a large animated FBX — viewport stays responsive with UV Editor closed
- [ ] Open UV Editor after import — layout + context islands populate correctly
- [ ] Custom palette users — menu checkbox still reflects saved theme, no startup repaint stall

## Summary by CodeRabbit

* **New Features**
  * The UV Editor now updates more efficiently in the background and resumes automatically when opened.
  * Several panels and tools now load only when needed, improving startup and display responsiveness.

* **Bug Fixes**
  * Hidden UV Editor panels no longer trigger unnecessary rebuilds or updates.
  * Theme changes now avoid extra UI work, helping prevent freezes.
  * Docked views are more reliable when shown or activated.

### DIFF
--- a/src/AssetBrowserController.cpp
+++ b/src/AssetBrowserController.cpp
@@ -8,6 +8,7 @@
 #include <QFileInfo>
 #include <QSettings>
 #include <QStandardPaths>
+#include <QTimer>
 
 AssetBrowserController* AssetBrowserController::m_pSingleton = nullptr;
 
--- a/src/UVEditorController.cpp
+++ b/src/UVEditorController.cpp
@@ -70,7 +70,14 @@ void UVEditorController::setPanelActive(bool active)
     if (m_panelActive == active)
         return;
     m_panelActive = active;
-    if (m_panelActive && m_refreshPending) {
+    if (!m_panelActive) {
+        if (m_refreshTimer->isActive()) {
+            m_refreshTimer->stop();
+            m_refreshPending = true;
+        }
+        return;
+    }
+    if (m_refreshPending) {
         m_refreshPending = false;
         refresh();
     }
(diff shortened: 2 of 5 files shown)


fernandotonon:
Addressed CodeRabbit review in `6634066`:

| Comment | Fix |
|---------|-----|
| Theme menu out of sync with `ThemeManager` precedence | Checkbox sync now uses the same `appearanceTheme` / `palette` / `custom` resolution as `ThemeManager::applySavedThemeFromSettings()` (still no second `setPalette()` call) |
| Asset Browser `importMeshRequested` dropped before dock shown | Signal wiring restored at startup; only QML `setSource` stays lazy on visibility |
| UV debounce timer fires after dock hidden | `setPanelActive(false)` stops the timer and marks refresh pending |
| AI Chat open paths missing breadcrumbs | Added `ui.action` breadcrumbs on toolbar + menu entry points |
| Hidden-panel test order-dependent | Fresh controller instance; assert `setPanelActive(true)` flushes pending refresh without explicit `refresh()` |

Latest CI run (28274082203) is green.

Issue #410, opened 2026-05-06
AI: Mesh segmentation / part labeling (ONNX)


fernandotonon:
Parent epic: #397

## Goal

Predict semantic part labels (head, torso, arm, leg, etc.) for a mesh using **PointNet++** or a similar small ONNX model. Powers smarter selection assist, auto-rigging hints, and per-part material assignment.

## Scope

- ONNX-exported PointNet++ trained on a part-segmentation dataset (ShapeNet-Part or similar).
- CLI: `qtmesh segment model.fbx --json` outputs vertex/face → label map.
- GUI: Edit Mode action "Select by part" — picks all faces sharing a predicted label.
- MCP: `segment_mesh` tool.
- Feed labels into the auto-rigger (#407/#408) as priors for skeleton placement.

## Acceptance Criteria

- [ ] Humanoid test mesh produces head/torso/limb labels with reasonable accuracy.
- [ ] "Select by part" populates the existing selection set.
- [ ] CLI JSON output is documented and stable.
- [ ] Sentry breadcrumb `ai.assist.segment` recorded.

## Dependencies

- ONNX Runtime (#404).

## Effort

~7 days.

PR #773, opened 2026-06-28
fix(#410): segmentation bone-side detection for non-Mixamo rigs

fernandotonon:
## Summary

Mining CC0 rigs (Khronos CesiumMan / RiggedFigure) for segmentation training data (#410) surfaced two gaps in `MeshSegmenter::partForBoneName` that affect any non-Mixamo standard rig:

1. **Side detection missed delimited side tokens.** The common `arm_joint_R_1` / `leg_joint_L__4_` naming puts the side letter between separators and a trailing index. `normaliseBoneName` strips separators → `armjointr1`, which ends in a digit, so the old `endsWith('l'/'r')` check failed and the joint fell through to **Torso** (one whole side of the body lost). New `boneSideFromRaw()` scans the raw name for an `l`/`r` token delimited by a separator/start and followed by a separator/digit/end — catches `_R_1`, `.l`, `-r`. It's the primary signal; the normalised heuristic stays as fallback.
2. **`torso`/`body` weren't torso keywords.** `torso_joint_N` bones resolved to Unknown. Added `torso`, `body`, `waist`, `sternum`.

## Impact

On the mined CC0 rigs, full resolution with **symmetric** L/R arms+legs (was missing one side entirely):

| mesh | before | after |
|------|--------|-------|
| CesiumMan | 3039/3273, no L-arm/R-leg | **3273/3273**, all 6 parts |
| RiggedFigure | 281/365, legs only | **365/365**, all 6 parts |

Helps Mixamo and any standard rig too (the rig-prior "Select by Part" path + the training-data miner both improve).

## Tests

Added `BoneSideFromDelimitedToken` (the `_R_1`/`.L`/`-r` cases + the negative case that a numeric-suffixed neck/torso joint is not dragged to a side) and torso-keyword assertions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bone-side detection for character meshes, correctly recognizing left/right markers in common naming patterns such as separated tokens and trailing numeric suffixes.
  * Reduced misclassification of center-body parts, including spine, neck, and torso-related joints, so they are more reliably grouped as central.
  * Added test coverage for mixed naming styles to help prevent future regressions.

### DIFF
--- a/src/MeshSegmenter.cpp
+++ b/src/MeshSegmenter.cpp
@@ -32,16 +32,26 @@ const char* const kPartIds[] = {
     "unknown", "head", "torso", "left_arm", "right_arm", "left_leg", "right_leg",
 };
 
-// Normalise a bone name for matching: lowercase, strip a leading
-// "mixamorig[N]:" / "bip01 " / "def_"-style prefix, drop separators so
-// "LeftArm" / "L_Arm" / "arm.l" / "DEF-upper_arm.L" compare alike.
-QString normaliseBoneName(const QString& raw)
+// Strip non-semantic prefixes (lowercased): a "mixamorig[N]:" / namespace
+// colon prefix and a leading "def_"/"org-"/"mch-"/"ctrl-" rig prefix. Keeps
+// separators intact (delimited-token side detection needs them). A namespace
+// like "LeftRig:Spine1" MUST be dropped here, or its "left" leaks into side
+// detection and a centre bone loses its torso mapping.
+QString stripBonePrefixes(const QString& raw)
 {
     QString s = raw.toLower();
     const int colon = s.lastIndexOf(':');           // mixamorig:LeftArm → leftarm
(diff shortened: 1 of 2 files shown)


fernandotonon:
Thanks @coderabbitai — fixed in 5fbfdcb. You're right: `boneSideFromRaw` scanned the namespace prefix, so `LeftRig:Spine1` leaked `'l'` and skipped the torso branch. Factored `stripBonePrefixes()` out of `normaliseBoneName()` (drops the namespace colon + def-/org-/mch-/ctrl- prefixes but KEEPS separators so delimited side tokens survive) and feed that to `boneSideFromRaw()`. A real side token after the namespace (`LeftRig:arm_R_1` → RightArm) still resolves. Added `BoneNamespacePrefixDoesNotLeakSide` covering both the leak and the still-resolves cases.

PR #771, opened 2026-06-27
feat(#410): AI mesh part segmentation (ONNX PointNet++ + geometric fallback)

fernandotonon:
Implements **issue #410 — AI: Mesh segmentation / part labeling (ONNX)**. Follows the epic #397 AI-assist pattern (#404/#408/#409). The **fourth ONNX consumer**.

## What it does
Predicts a semantic part label (head / torso / left+right arm / left+right leg) per vertex and per face of a mesh. Powers Edit-Mode "Select by part", per-part material assignment, and auto-rig priors. Surfaced on CLI, MCP, and the Edit Mode panel.

## Geometric fallback is first-class
`MeshSegmenter::segmentGeometric` — connected-component islands (union-find) + an up-axis/lateral spatial heuristic (top→head, lower→legs, mid-sides→arms, centre→torso), overridable per-vertex by rig bone-proximity hints — is always compiled (Ogre-free, no ONNX) and used automatically when the build lacks ONNX, the model is missing/un-downloadable, or inference fails. `Result::usedModel` / `fallbackReason` report which path ran. Verified on a Mixamo humanoid: head/torso/arms/legs all populated with sensible proportions.

## ONNX path
PointNet++-style: normalise → deterministic point sample → `[1,N,3]` → per-point argmax over part channels → scatter labels back to all vertices by nearest sampled point. Runtime I/O-name discovery, channels-first/last handling, CoreML EP. `ensureModelBlocking()` downloads `meshseg.onnx` to `AppData/ai_models/segment/` (overrides `QTMESH_SEGMENT_MODEL_BASE_URL` / `QSettings ai/segmentModelBaseUrl`; offline guard `QTMESH_SEGMENT_NO_DOWNLOAD`; non-ONNX `#ifndef` guard).

## Surfaces (CLI / MCP / GUI)
- **CLI:** `qtmesh segment <file> [--json] [--no-model] [--up-axis x|y|z]` — per-part vertex/face counts (text) or full vertex/face→label arrays + per-part summary (stable JSON schema).
- **MCP:** `segment_mesh` tool — `{entity_name?, no_model?}`, returns per-part vertex counts + which path ran (registered heavy).
- **GUI:** Edit Mode → **"Select by Part (AI)"** — selects all faces matching the selected face's part (or the largest part if none selected); pushes via `selectFace` so the existing highlight refreshes.

## Model (trained + hosted)
[...]

PR #774, opened 2026-06-28
fix(#410): segmentation works in-app — sampling bug + flat model (3.13.1)

fernandotonon:
## Problem

The mesh part segmentation shipped in 3.13.0 (#771) was effectively broken **in the app** — it scored ≈0.33 per-vertex accuracy even on a correctly-oriented rigged mesh (the model itself was fine: 0.98 in Python). Two independent root causes, found by feeding the C++ path's exact points back through Python ORT:

### 1. Inverted point sampling (`MeshSegmenter::predict`)
For meshes with fewer vertices than `samplePoints` (4096 — i.e. **most** meshes), the sampling loop drew **every** point at random instead of covering verts `0..vc-1` in order. The nearest-point scatter then only labelled the randomly-hit verts (~63% coverage) and left the rest at the `Torso` default. Fixed: the first `min(N, vc)` points are verts in order (1:1 coverage), and `N` is capped at the vertex count so small meshes aren't padded with noisy duplicates.

### 2. In-graph kNN incompatible with the linked ONNX Runtime
The model used an in-graph kNN (`cdist`/`topk`/`gather`). Those ops did **not** execute correctly in the app's ONNX Runtime build — identical model + input gave 0.98 in Python ORT but 0.33 in the C++ app. Reworked the exported model to a **flat PointNet segmentation network** (per-point MLP → global max-pool → per-point head) using only `Linear`/`GELU`/`max`, which every ORT build runs identically.

**Verified in the C++ app:** 0.33 → **0.99** on a correctly-oriented rigged mesh.

## Model — meshseg v1.2.0 (flat)

Retrained on synthetic + **~89 CC0 rigged characters** (Quaternius CC0, Khronos glTF-Sample-Assets, ToxSam/Polygonal-Mind `xyz` CC0). Hosted on the HF repos (combined + dedicated). Training script gains: flat arch, `--real-data`/`--real-weight` rebalance with per-source (synth vs real) val reporting, and full 360° yaw augmentation.

**Known limitation (documented):** left/right limb assignment is inherently orientation-dependent — best on upright, front-facing meshes; a backward-facing mesh may mislabel L/R. **Rigged meshes are unaffected** — they use the exact rig-prior path, not the model.

## Version

Bumps to **3.13.1** (3.13.0 shipped the broken model).

## Summary by CodeRabbit

* **New Features**
  * Mesh segmentation now labels small meshes more reliably, reducing missing or default labels.
  * Training and validation workflows now better distinguish real and synthetic data, with support for weighting real samples more heavily.

* **Bug Fixes**
  * Improved point sampling so it no longer exceeds the number of vertices.
  * Updated ONNX export behavior and reporting for more consistent results.

* **Documentation**
  * Bumped release and CI examples to version **3.13.1** throughout the docs and website fallback settings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ cmake_minimum_required(VERSION 3.24.0)
 cmake_policy(SET CMP0005 NEW)
 cmake_policy(SET CMP0048 NEW) # manages project version
 
-project(QtMeshEditor VERSION 3.13.0 LANGUAGES C CXX)
+project(QtMeshEditor VERSION 3.13.1 LANGUAGES C CXX)
 message(STATUS "Building QtMeshEditor version ${PROJECT_VERSION}")
 
 set(QTMESHEDITOR_VERSION_STRING "\"${PROJECT_VERSION}\"")
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Available on the [GitHub Actions Marketplace](https://github.com/marketplace/act
 **Versioning**
 
 - **Always follow the latest GitHub release** — use the Marketplace floating tag `fernandotonon/QtMeshEditor@v1` (same pattern as the [Marketplace example](https://github.com/marketplace/actions/qtmesheditor)). The composite action defaults to `image-tag: latest`, so the Docker CLI tracks the newest published `ghcr.io/fernandotonon/qtmesh` image.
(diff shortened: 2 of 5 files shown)


fernandotonon:
Closing — after extensive experimentation (8 model iterations + connectivity/smoothing/auto-rig + a Mixamo-vs-CC0 training comparison), we concluded the segmentation rework wasn't a clear improvement over the shipped model, and reverting to keep the known-good state. Key findings documented for the future: the in-app model load path is nested (QtMeshEditor/QtMeshEditor/), the sampling loop had an inversion bug worth revisiting, and (settled empirically) Mixamo training data does NOT beat CC0/synthetic. Parking this rather than shipping a regression.

PR #795, opened 2026-07-03
fix: PBR generation renders the surface too dark (roughness/metallic FFP wiring)

fernandotonon:
## Summary

Generating PBR maps (Material Editor "Generate PBR maps" or the image-to-3D PBR step) made the model render **dark — "like the lights are off on the model," while other meshes lit normally.** Export→reload of the same mesh looked correct, which localized both bugs to the live in-session material wiring.

Two independent root causes, both fixed:

### 1. Roughness/metallic darkened the diffuse (FFP wiring)

The FFP fallback wired `roughness` as `LBX_MODULATE_X2` and `metallic` as `LBX_ADD_SIGNED` against the running colour — but those are BRDF specular-lobe inputs, not colour channels. A typical generated roughness map (mean ≈ 0.44) multiplied the visible colour by `roughness × 2 ≈ 0.87`, tinting the surface darker. **Fix:** make roughness/metallic inert in the FFP path (`LBX_SOURCE1`, non-FFP), like normal maps; the real contribution comes from the Cook-Torrance SRS when PBR-tagged with IBL. (`RTShaderHelper::wirePbrSlotsForFFP` + `MaterialPresetLibrary::configurePbrSlots`.)

### 2. RTSS normal map without tangents → unlit surface (the main "lights off" symptom)

The live paths wired the RTSS `SRS_NORMALMAP` sub-render-state on a mesh with **no per-vertex tangents**. Ogre's normal-map shader needs tangents to build the tangent-space basis; without them the basis is degenerate, the perturbed normal collapses, `N·L → ~0`, and the surface renders as if unlit. The **mesh importer already builds tangents before wiring normal maps** (`applyNormalMapsToEntity`, *"Build tangent vectors … required by RTSS normal mapping"*) — which is exactly why export→reload looked fine but live generation didn't. **Fix:** route both live paths through that same tested importer routine:
- image-to-3D (`MeshGenBuilder`): defer normal-map wiring until the entity exists, then call `applyNormalMapsToEntity` (builds tangents when UVs exist, then wires SRS_NORMALMAP).
- Material Editor (`generatePbrFromDiffuse`): after binding the maps, run `applyNormalMapsToEntity` on every scene entity using the material.

## Verification

- Roughness tint: lit-face brightness 126 → 150 on the live bind path; regression test asserts roughness/metallic resolve to `LBX_SOURCE1`.
- Normal-map/tangent darkening: reproduced on a primitive sphere (brightness 104 → 81 after the old raw `applyNormalMap`); after the fix the image-to-3D generated mesh lights the same as an export→reload (verified in the viewport).

## Paths covered

Both `wirePbrSlotsForFFP` (Material Editor / image-to-3D / CLI / MCP) and `applyNormalMapsToEntity` routing (the two live viewport paths). Export/CLI/MCP paths were already correct.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PBR metallic and roughness inputs so they no longer darken the material in fixed-function fallback mode.
  * Improved normal map handling for generated and imported materials so lighting appears correctly on live meshes.

* **Tests**
  * Added regression coverage for the updated PBR fallback behavior.

### DIFF
--- a/src/ImageTo3D/MeshGenBuilder.cpp
+++ b/src/ImageTo3D/MeshGenBuilder.cpp
@@ -2,6 +2,7 @@
 
 #include "Manager.h"
 #include "AIAssistManager.h"   // #404 PBR map synthesis (normal + roughness)
+#include "MeshImporterExporter.h"  // applyNormalMapsToEntity (tangents + RTSS)
 #include "RTShaderHelper.h"    // canonical-slot FFP wiring + RTSS normal map
 
 #include <QDateTime>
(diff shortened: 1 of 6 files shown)


fernandotonon:
@chatgpt-codex-connector re: specular-glossiness slot wiring — fixed in 6e6b65f. `configurePbrSlots` now takes the workflow: metallic-roughness keeps the inert `LBX_SOURCE1` pass-through (real response via Cook-Torrance when IBL is present), but specular-glossiness — which stays on the FFP path since `applyPbrIfTagged` skips it — retains the legacy `ADD_SIGNED` (specular) / `MODULATE_X2` (glossiness) ops so those maps still affect the render.

Issue #805, opened 2026-07-04
Image-to-3D: MV-Adapter (Apache-2.0) as a multi-view texture upgrade for TripoSG


fernandotonon:
# Image-to-3D: MV-Adapter (Apache-2.0) as a multi-view texture upgrade for TripoSG

## Motivation

TripoSG (#764) gives us the best local single-image **geometry**, but it's geometry-only. Today its colour comes from (a) projecting the input photo onto the visible **front** and (b) inferring the **back** from TripoSR's colour field. The front is photo-accurate; the back is approximate and can be blotchy where the two models' scales disagree.

**MV-Adapter** (huanngzh/MV-Adapter, *"Multi-view Consistent Image Generation Made Easy"*) generates a set of **consistent multi-view images** conditioned on geometry — the same technique Meshy/Tripo use before baking. Feeding those views into our existing `MultiViewTextureBaker` would give a genuinely good, consistent texture on **all** sides, not just the front.

This is the tracked follow-up from the #764 texture work (the `generate3d` "Generate texture (AI)" option currently uses a photo-front + SD-txt2img-back multi-view bake; MV-Adapter replaces the back-generation with proper multi-view-consistent generation).

## License due-diligence (vetted — GO)

- **Code:** Apache-2.0 (GitHub `LICENSE`, verbatim Apache 2.0).
- **Adapter weights:** Apache-2.0 (HF `huanngzh/mv-adapter` model card). Variants: `t2mv`/`i2mv`/`tg2mv`/`ig2mv` for SD2.1 and SDXL bases; the geometry-guided `tg2mv`/`ig2mv` SDXL checkpoints are the texturing ones.
- **Training data:** ~70K filtered **Objaverse** subset (ODC-BY-1.0) + **Cap3D** captions (permissive). No non-commercial dataset; no UltraSharp/ambientCG-style provenance discrepancy.
- **Base model (the caveat):** MV-Adapter is an *adapter* — it needs an **SD2.1 or SDXL** base, both **CreativeML Open RAIL++-M**. This is commercial-and-redistribution-OK with ethical use-restrictions, and is the **same RAIL license family we already ship via the SD1.5 `stable-diffusion.cpp` path** — no new licensing category. NOT in any rejected bucket (SF3D/Zero123 NC, Hunyuan3D EU-excluded, GPL deps).
[...]

PR #794, opened 2026-07-02
Image→3D: TripoSG backend — 1.5B rectified-flow geometry (MIT), selectable next to TripoSR

fernandotonon:
## What

Adds **TripoSG** ([VAST-AI-Research](https://github.com/VAST-AI-Research/TripoSG), SIGGRAPH 2025 — **MIT code + MIT weights**, verified) as a second image→3D backend, selectable next to TripoSR on every surface. TripoSG is a 1.5B rectified-flow DiT over an SDF VAE with reported geometry quality ≈ commercial Tripo 2.0 (Normal-FID 5.81 vs ~20 for TripoSR-class LRMs) — the "best geometry" tier from the `docs/IMAGE_TO_3D_QUALITY.md` roadmap.

## How it runs (no torch at runtime — 7th ONNX consumer)

Four exported graphs (`scripts/export-triposg-onnx.py`, offline dev tool; measured contract in `docs/TRIPOSG_EXPORT_NOTES.md`, extracted from upstream source + HF configs):

1. **DINOv2-224 image encoder** — preprocessing (ImageNet mean/std) baked into the graph; CFG's unconditional embedding is zeros (upstream `zeros_like`).
2. **DiT step** — one denoising evaluation; the **rectified-flow Euler loop runs in C++** (same hand-rolled-loop stance as UniRig's KV-cache decode): σᵢ = 1 − i/N, timestep = 1000·σ, update `x += (σᵢ − σᵢ₊₁)·v` — note this sign is the **opposite** of stock diffusers FlowMatchEuler (TripoSG's custom scheduler). CFG as two B=1 calls, guidance 7.0, user-tunable steps (default 25; 50 = reference, 10 = preview). fp32 ships as `.onnx` + `.onnx.data` (>2 GB); `--quality int8` maps to a single-file int8 DiT tier.
3. **VAE latent kv-cache** — the 16-block latent self-attention stack, run **once** per generation.
4. **Per-point field decoder** — (kv_cache, points) → inside-positive field, iso 0, bounds ±1.005; tiled in chunks and extracted with the same native MarchingCubes, then the existing Taubin/reprojection polish.

TripoSG is **geometry-only** (no colour decoder): the texture bake/PBR/upscale stages stay TripoSR-only and their GUI checkboxes auto-disable; background removal composites over **white** per TripoSG's reference pipeline (vs TripoSR's gray-128). Deterministic seed → same image + params = same mesh.

## Surfaces

[...]

PR #827, opened 2026-07-09
feat(#805): MV-Adapter slice 1 — 4-view AI texture bake for image→3D meshes

fernandotonon:
## MV-Adapter slice 1 (#805): 4-view AI texture bake + image→3D texture fixes

Builds on the TripoSG image→3D backend (#794, shipped in 3.18.0) to give the geometry-only meshes a real, image-derived texture — and fixes the chain of bugs that kept that texture from actually reaching the screen.

### What's new
- **4-view AI texture bake** (MV-Adapter #805 slice 1): front/back/left/right depth-ControlNet views (its 6-view orthographic layout minus the two poles), projected onto the mesh's UVs and blended. A clear step up from the old front+back (which left the sides to stretch/blur) without the ~3× cost of all six.
- **Describe-then-generate**: a local **SmolVLM** (Apache-2.0) vision model captions the input image; the caption drives every view's SD generation, so front/back/sides are stylistically consistent (shared caption + locked seed) with no photo-projection registration artifact.
- **Unwrap up front**: the mesh is degenerate-cleaned, decimated to a tri budget, and UV-unwrapped *before* the SD views run (not after), so every view and the final bake share the same geometry — and the long unwrap no longer looks like a freeze after generation.

### Bug fixes (this round — reported during hands-on testing)
1. **Texture never rendered on TripoSG meshes.** The geometry-only `MeshGen/NeutralClay` material has no texture unit, so the successfully-baked diffuse had nowhere to bind and the mesh stayed clay. Now: create a named `diffuse_map` TUS (white base tint), clone the shared clay material per-entity so one bake doesn't tint every generated mesh, and — the crucial one — remove the stale RTSS technique directly off the material so it regenerates from the FFP pass (a cloned material's RTSS technique isn't tracked by `removeAllShaderBasedTechniques`, which is why only a manual material Apply "fixed" it).
2. **Status frozen at "view 4/4 — step 12/12".** `sdTextureGenerated` was emitted per view, prematurely marking the bake done on view 0; now emitted once after the real apply.
3. **SmolVLM captioned a person as "a fluffy brown and white rabbit".** The prompt contained a worked example the tiny 500M model parroted verbatim. Removed the example, rewrote the prompt to name the subject first then surface detail (and ignore the background), fixed the chat template to match the GGUF's embedded Jinja exactly, and serialised `caption()` (concurrent caption threads corrupted llama.cpp global state).

### Models
- **SmolVLM-500M-Instruct** (GGUF + mmproj) — Apache-2.0 — hosted on the `fernandotonon/QtMeshEditor-models` HF repo under `caption/`, downloads on first use. Verified present (HTTP 302).
- No new model uploads required in this round; only code/prompt changes.

### Testing
- Verified end-to-end on macOS: image → TripoSG geometry → 4-view bake → texture renders on the mesh without a manual Apply; caption correctly describes people, animals, and objects (identity + colours/materials).

## Summary by CodeRabbit

* **New Features**
  * Improved AI image-to-3D texture generation by sampling more view angles, leading to better multi-view texture results.
* **Bug Fixes**
  * Made local image captioning more reliable under concurrent use.
  * Refined image descriptions to focus more accurately on the main subject.
  * Fixed generated textures from affecting other meshes when materials are shared.
  * Improved texture binding for models that were missing a proper diffuse texture slot.

### DIFF
--- a/qml/PropertiesPanel.qml
+++ b/qml/PropertiesPanel.qml
@@ -2111,9 +2111,17 @@ Rectangle {
                         // image was picked (ready by now — no UI-blocking
                         // captioning here). Empty falls back to a neutral prompt
                         // inside generateMeshTextureMultiView.
+                        //
+                        // Views: front/back/left/right (MV-Adapter #805 slice 1
+                        // — its 6-view orthographic layout minus the two poles).
+                        // 4 equatorial views give full horizontal coverage +
+                        // seam overlap for the baker's cross-view blend, a clear
+                        // step up from the old 2 (front+back left the sides to
+                        // stretch/blur) without the ~3× cost of all 6. The two
+                        // poles (top/bottom) rarely help typical subjects.
                         MaterialEditorQML.generateMeshTextureMultiView(
                             MeshGenController.caption, 512, 512, 0.9,
-                            ["front", "back"],
+                            ["front", "back", "left", "right"],
                             "",                       // no photo pinning
                             mgPbr.checked)
                     }
--- a/src/ImageTo3D/ImageCaptioner.cpp
+++ b/src/ImageTo3D/ImageCaptioner.cpp
@@ -16,6 +16,7 @@
 #include <mtmd.h>
 #include <mtmd-helper.h>
 
+#include <mutex>
 #include <string>
 #include <vector>
 #endif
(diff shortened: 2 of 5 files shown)


fernandotonon:
Addressed the reviews:

**Codex P2 (vertex-colour tracking on MeshGen materials)** — valid, fixed in bf5d1cd. When repurposing a `MeshGen/*` no-texture material we now call `pass->setVertexColourTracking(Ogre::TVC_NONE)`, so the baked AI diffuse replaces the decoder vertex colours instead of being multiplied by them. No-op for `NeutralClay` (never enabled tracking); the fix matters for the TripoSR + AI-texture `MeshGen/VertexColor` case.

**CodeRabbit (move captioning off the UI thread in `generateMeshTextureMultiView`)** — the inline caption branch (`ImageCaptioner::ensureModelBlocking()` + `caption()`) only runs when `prompt.isEmpty() && !frontPhotoPath.isEmpty()`. In the shipping image→3D flow, `MeshGenController` captions the image on a background thread the moment it's selected and QML passes that pre-computed caption as the `prompt` (with `frontPhotoPath = ""`), so this branch **does not execute** — the UI-blocking path is already avoided by design. The branch is a fallback for a hypothetical direct caller that supplies a photo with no prompt; restructuring the synchronous multi-view kickoff into a worker for a dead-in-practice path would add real complexity for no shipping benefit, so leaving it. Happy to revisit if a non-image→3D caller starts using the photo-only form.

PR #829, opened 2026-07-10
feat(skinning): Skinning v2 — geodesic voxel binding default + weight post-passes (#819 Slices A+B)

fernandotonon:
Implements Slices A and B of #819, with full GUI/CLI/MCP surface parity and the Slice-C enum plumbing.

## What's in

**Slice A — Geodesic Voxel Binding (`src/GeodesicVoxelBind.{h,cpp}`, new default)**
The method Maya ships as "Geodesic Voxel" bind (Dionne & de Lasa, SCA 2013), implemented natively with zero new dependencies:
- Surface voxelization via Akenine-Möller triangle/box SAT at `voxelResolution` (default 64, max 256)
- Exterior flood-fill classification — closes holes at voxel resolution, so non-watertight / self-intersecting / multi-component meshes work
- 3D-DDA bone-segment rasterization; bones outside the solid snap to the nearest solid voxel within 4.5 voxels (world-distance-checked) or get no seeds and are reported
- One multi-source Dijkstra over interior+surface voxels (26-connectivity), each voxel keeping its best K=8 (bone, distance) pairs — one pass total, not one per bone
- Per-vertex `(1/d)^falloff` weighting reusing the existing `SkinWeightsOptions` semantics unchanged

Cross-limb bleed is impossible by construction: distances travel through the volume, so a hand near a thigh can never pick up leg weights.

**Slice B — Weight post-pass pipeline (`src/SkinWeightsPost.{h,cpp}`, applies to every algorithm)**
- Laplacian relaxation over the vertex adjacency graph (default 3 iterations, `--smooth-iterations`, 0 = off). Merge-mode (`--merge`) manual weights act as Dirichlet constraints — they shape the blend at the boundary but are never modified
- Prune < 0.01 + top-K + renormalize (partition of unity guaranteed; largest weight always survives)
- Bleed report metric: fraction of committed weights not geodesically local (0 for GVB by construction)

**Whole-mesh single-pass compute**: `computeAndApply` now gathers every vertex-data owner (shared + per-submesh) into one combined vertex/index set and computes once — matching the paper (voxelize the whole character). This was load-bearing: per-submesh grids stranded bones outside accessory AABBs (verified on a 39-submesh character where per-owner grids reported all 119 bones seedless).

**Algorithm enum + fallbacks**
- `Algorithm { InverseDistance, GeodesicVoxel (default), UniRigML }`
- Volume-less input (planes, cloth, billboards) → automatic InverseDistance fallback, never fails
[...]

Issue #819, opened 2026-07-06
AI: Skinning v2 — Mixamo-quality auto skinning (GVB default, UniRig skin head, weight post-passes, DQS)


fernandotonon:
Parent epic: #818 (Track B companion — Skinning v2) · Upgrades #402 (`SkinWeights`, closest-point heuristic) · Pairs with #818-B1 (UniRig arbitrary categories: rigid/vehicle binding stays there; this issue owns deformation quality)

# AI: Skinning v2 — Mixamo-quality auto skinning (GVB default · UniRig skin head · weight post-passes · DQS)

## Problem

`SkinWeights` (`src/SkinWeights.{h,cpp}`, #402) implements the classic closest-point-on-bone smooth bind: per vertex, inverse **Euclidean** distance to each bone segment, falloff exponent, top-K, normalize (`SkinWeights.h:32-44`). Euclidean distance travels *through* the body's volume, so this method has a hard quality ceiling no tuning fixes:

- A hand posed near a thigh picks up leg weights; inner thighs share weights across legs; fingers grab each other. `maxInfluenceDistance` (`SkinWeights.h:59-63`) only caps range — it cannot express "near in space but far along the body".
- Joint falloffs are distance shells, not anatomy-following gradients → visible creasing/collapse at elbows/knees/shoulders on 90° bends.
- The header's own escape plan — libigl BBW — stays blocked: TetGen is GPL (`SkinWeights.h:23-31`). This issue implements the two paths that beat BBW *in production practice* without the license problem.

Target bar ("Mixamo level", made testable in Slice E): smooth falloffs across every major joint at 90° bends, zero cross-limb bleed, ≤ 4 influences/vertex, works on non-manifold / non-watertight / multi-component meshes (auto-generated TripoSG output, downloaded FBX).

## Architecture (all slices)

New algorithms slot into the existing enum and pure-data contract:

```cpp
enum class Algorithm {
    InverseDistance,   // existing heuristic (kept: fallback + planes/cloth)
    GeodesicVoxel,     // Slice A — new DEFAULT
    UniRigML,          // Slice C — ENABLE_ONNX, auto-falls-back to GeodesicVoxel
};
```

[...]

Issue #408, opened 2026-05-06
AI: RigNet auto-rigging (ONNX, ML)


fernandotonon:
Parent epic: #397

## Goal

Higher-quality auto-rigging than Pinocchio using **RigNet** (2020, ONNX-exportable). Better handling of arbitrary topology, non-humanoid shapes, and unusual proportions.

## Scope

- Reuse the ONNX Runtime infrastructure landed by #404 (PBR synth).
- Vendor a working RigNet ONNX export (will need to verify availability — fallback: do the export from the published PyTorch checkpoint as a one-time tooling step).
- Model size: ~50 MB; download via `ModelDownloader` on first use.
- CLI: `qtmesh rig model.fbx --algo rignet -o rigged.fbx` (default: pinocchio for offline reliability).
- GUI: skeleton-prediction algorithm dropdown alongside #407.
- MCP: extend the `auto_rig` tool with an `algo` param.
- Pipe into #402 (libigl skin weights) the same way as #407.

## Acceptance Criteria

- [ ] Works on humanoid, quadruped, and at least one non-standard shape (e.g. chair-with-legs).
- [ ] Output skeleton hierarchy matches what a reasonable artist would author.
- [ ] First-run model download flow with progress and offline failure messaging.
- [ ] CLI/MCP/GUI parity.
- [ ] Sentry breadcrumb `ai.assist.auto_rig` with `algo: rignet` tag.

## Dependencies

- ONNX Runtime infra (#404).
- Pinocchio task (#407) for the rigging plumbing.

## Effort

~10–12 days.

PR #830, opened 2026-07-10
feat(skinning): Skinning v2 — DQS display toggle, evaluation suite, Slice-C decision (#819 Slices C/D/E)

fernandotonon:
Completes the remaining slices of #819 on top of the merged Slices A+B (PR #829).

## Slice D — Dual-quaternion skinning display toggle

No weight map fixes linear-blend volume collapse on twists — that needs a different blend at render time.

- **`SkinningDisplay`** (`src/SkinningDisplay.{h,cpp}`): per-entity toggle via Ogre's RTSS `HardwareSkinningFactory`. Dual Quaternion = `prepareEntityForSkinning(ST_DUAL_QUATERNION, antipodality-corrected)` + technique invalidation; Linear = erase the material's `HS_SRS_DATA` imprint (mirrors Ogre's file-local constant, stable since 1.8) and regenerate. Mode is tracked on the entity's `UserObjectBindings`.
- The factory + a **dormant** template SRS are registered in `RTShaderHelper::initialize` (Ogre ShaderSystem-sample pattern — plain materials are unaffected) and unregistered + deleted on shutdown. Bone cap 96: entities above it stay safely on the default path (verified: 119-bone bandit renders unchanged).
- Surfaces: **"Display: Linear | Dual Quaternion"** row in the Animation-mode Skinning section, MCP **`set_skinning_display`** (`mode` enum), `SkinWeightsController::skinningDisplayMode/setSkinningDisplayMode`, Sentry `render.skinning`.
- **Display only** — exporters consume the unchanged vertex weights (documented in tooltips + MCP description + docs).
- Live-verified through the running app (load → toggle DQS → toggle Linear → renders correctly both ways). One pre-existing finding while verifying: quitting via AppleScript aborts (exit 134) — reproduces with the new factory disabled, i.e. **not** introduced here.

## Slice E — Metrics, acceptance suite, Mixamo protocol

- **`SkinMetrics`** (pure-data, unit-tested): influence histogram, Laplacian weight-smoothness energy, LBS deform (`rotationAbout` + `deformLBS`), signed mesh volume.
- **`SkinEvaluate`**: extracts an entity's **existing** weights (whoever authored them) and reports metrics — including the geodesic bleed of existing weights against a fresh `GeodesicVoxelBind` field — plus a reference comparison with position-matched vertices and name-matched bones. Equidistant duplicate vertices (seams, hand-on-prop contact points) tie-break on minimum weight difference, so self-compare is **exactly 0** (verified on the 90k-vert bandit; the naive matcher reported spurious 1.45 max-diffs at contact points).
- **CLI**: `qtmesh skin <file> --evaluate [--voxel-res N] [--json]` and `qtmesh skin <file> --compare <reference> [--json]`.
- **Acceptance fixtures** (hermetic, procedural — no binary assets): 90° elbow-bend capsule volume ratio **≥ 0.9** (measured 0.911 with geodesic weights + Slice-B smoothing), two-limb proximity bleed **= 0** for geodesic vs **> 0.1** for the inverse-distance contrast, smoothing strictly reduces Laplacian energy, plus metric primitive tests.
- **`docs/SKINNING_QUALITY.md`**: metric definitions, thresholds, the manual Mixamo comparison protocol (no Adobe assets in-repo), and the env-gated regression test (`QTMESH_SKIN_OURS_FBX` / `QTMESH_SKIN_REF_FBX` — skipped in CI, suite has unconditional tests so it never goes skip-only).

## Slice C — UniRig skinning head: decision record

Per the issue's own decision gate ("also evaluate SkinTokens … if it exports cleaner, prefer it"):

- **UniRig's skin head is not ONNX-exportable as-is**: verified against upstream `src/model/unirig_skin.py` — the mesh geometry path runs through a **PTv3 backbone on spconv sparse convolutions** (no ONNX operator lowering) plus flash-attn MHA. The #408 skeleton export only worked because that stage never executes the PTv3 path. A faithful `skin.onnx` means re-implementing PTv3 densely — a research task.
- **SkinTokens/TokenRig** (VAST-AI, MIT code, FSQ-CVAE weight tokenizer + Qwen3-0.6B AR transformer, **no spconv**) is the preferred ML-skinning follow-up — exportable with the #408 KV-cache decoder pattern, but a full new integration rather than a drop-in head. Weights-license confirmation is part of that follow-up.
- Recorded in `THIRD_PARTY_AI_MODELS.md`; `Algorithm::UniRigML` keeps its clean, tested fallback to GeodesicVoxel on every surface meanwhile, so the ML path can light up later without changing any app surface.

## Verification

- All pure-data fixtures verified via the standalone harness (macOS can't run Ogre-backed UnitTests locally; Linux CI runs the full suite): elbow ratio 0.911, bleed contrasts, smoothing energy reduction.
- `--evaluate` on the 90k-vert bandit: avg 3.44 influences, bleed 0.028, smoothness 0.0124. `--compare` self-test: 90,573/90,573 matched, mean/max L1 diff 0.0000.
- DQS toggle exercised live via MCP on two real assets, screenshots inspected.
- App + UnitTests build clean.

Tracks #819 (Slices C/D/E; A+B merged in #829).

## Summary by CodeRabbit

* **New Features**
  * Added Linear and Dual Quaternion skinning display modes in the Animation Mode tools (visual-only).
  * Added MCP tool support to switch the selected entity’s skinning display mode at runtime.
  * Added `qtmesh skin --evaluate` (quality metrics) and `qtmesh skin --compare <reference>` (per-vertex weight diffs) with JSON or text output.
* **Documentation**
  * Expanded skinning-quality guidance, acceptance workflow, and dual-quaternion display behavior.
  * Updated CLI and skinning decision records.
* **Tests**
  * Added end-to-end and unit coverage for display mode toggling, skin metrics, deformation/smoothing/bleed, and reference comparisons.


fernandotonon:
Both Codex findings were valid — fixed in the latest commit:

- **P1 (imprint on the technique)**: confirmed against Ogre v14.5.2 source (`imprintSkeletonData` stores on `getTechnique(0)->getUserObjectBindings()`, `preAddToRenderState` reads the same) — Linear mode now erases the technique binding, and the test's imprint probe checks the same place.
- **P2 (shared materials)**: `apply()` now stamps the tracked mode on every entity sharing one of the affected materials, so `current()` stays truthful for all of them; covered by the new `SharedMaterialEntitiesTrackTheSameMode` test. (Material uniquification was deliberately avoided — silently cloning materials would leak clone names into exports and the Material editor.)

Also fixed the CI failure: the test harness never runs `Manager::loadResources()`, so `SkinningDisplayTest` now initializes RTSS explicitly in SetUp (the `MaterialPresetLibrary_test` pattern).

fernandotonon:
CI is green — 2418ed0 fixes the unit-tests-linux failure.

Root cause: Ogre 14.x registers `HardwareSkinningFactory` as a ShaderGenerator **built-in** (`mBuiltinSRSFactories`). The Slice D shutdown deleted that built-in, and `ShaderGenerator::destroy()` then deleted it again — the double free segfaulted teardown of every suite running the full RTSS lifecycle (7 crashed suites → executed-count gate failure). Meanwhile the matching `initialize()` block was guarded by `!getSingletonPtr()`, which was always false, so the bone cap and template SRS were silently never applied.

Now per the Ogre DualQuaternion sample: use the built-in factory, set the bone cap, add the template hardware-skinning SRS to the generated scheme (deduped by type — `addTemplateSubRenderState` only dedupes by instance and tests re-run initialize), and leave the factory alone at shutdown. Side effect: the DQS display toggle's SRS actually reaches generated techniques now.

PR #904, opened 2026-07-14
t2m v2 (epic #837): twist transport, foot pinning, flow-matching model

fernandotonon:
Closes #857, closes #856, closes #840, closes #858 — the remaining slices of epic #837, combined into one PR (supersedes the stacked #886/#887/#888).

## 1. Twist transport in the direction retarget (#857)

The bind-referenced retarget (PR #843) aimed bone directions but dropped roll — gesture clips read flat. The twist is now decomposed on the source and recomposed **about the same pole the source decomposed about**:

```
Qbase = arc(dt_bind → d_ref) · Wt_bind        (once per bone)
Wt(f) = twist(θ·gain, ds) · arc(d_ref → ds) · Qbase
```

Recomposing from the target-bind pole (the issue's original formula) leaks roll into swing — measured +38% arm amplitude. With matched poles, per-frame relative motion is exactly the source's world delta conjugated into target axes. θ is unwrapped across frames (a ±180° pop would flip the roll once a gain ≠ 1 scales it), capped 150°; collars damped 0.5×.

**Harness:** Mixamo Walk self-parity **2.22° → 0.03°** mean joint delta (gate ≤1.6°); spine amplitude restored (abdomen 2.3°→9.0°); Quaternius Knight cross-rig 6.68°→5.36°, renders upright. Controlled A/Bs pin a single-clip eval library — the runtime take-pick is quality²-weighted RANDOM, so uncontrolled comparisons are noisy.

## 2. Foot-contact detection + IK pinning (#856)

Retargeted clips skate/float feet on rigs whose proportions differ from the source. New `FootContact` pure-data core (Ogre-free, 8 unit tests): leg-length-scaled contact detection (height band + horizontal-speed gate; a fast on-ground foot — skating — is correctly NOT a contact), analytic two-bone `solveKnee` (keeps segment lengths + the pose's own bend plane), 0→1→0 edge blends. `AnimationMerger::pinFeet` runs manual FK over the tracks (the live skeleton is never `apply()`'d), detects in the canonical frame, rewrites only thigh/shin/foot keyframes (foot keeps its original world orientation — no toe pop). Effectively idempotent.

**Surfaces:** ON by default in generation — CLI `--no-foot-pin` / standalone `--foot-pin --animation <n>`; MCP `foot_pin` arg + `pin_feet` tool; GUI "Pin feet" checkbox.

[...]

Issue #837, opened 2026-07-10
Epic: Text-to-motion v2 — permissive motion corpus + flow-matching model


fernandotonon:
Successor to #411 and the #793 decision record: no licensable pretrained text-to-motion model exists (HumanML3D/AMASS models are non-commercial; HY-Motion 1.0 is territory-restricted incl. its Outputs and derivatives-by-distillation), and our from-scratch CVAE's weakness is structural — a z=0 conditional-mean decoder averages modes into gentle motion (measured in the v4 training notes). Nobody in the open ecosystem has assembled a *permissive* motion corpus at scale — so we build one, then train the right architecture on it.

## Why this can work

- **Data**: thousands of permissively-licensed animated humanoid rigs exist across Sketchfab (API filters: CC0/CC-BY + animated + downloadable), OpenGameArt, and CC0 asset packs (Quaternius, Kenney, KayKit, Poly Pizza) — plus the full CMU corpus (~2,600 trials; we currently mine 47 clips) and SFU mocap (free incl. commercial, attribution). Asset names/tags provide free text labels ("zombie walk", "sword slash") — far richer than the current 13-action vocabulary.
- **Infrastructure we already own**: the 22-joint canonical skeleton + retarget (`applyMotionClip`, `canonicalIndexForBone`), library windowing/energy selection (`build-motion-library.py`), a C++ flow-matching sampler loop (TripoSG), and ONNX inference everywhere.
- **Architecture**: flow-matching/diffusion doesn't mode-average — it is the single change that separates MDM-era quality from our CVAE, independent of data scale.

## Slices

- **A — corpus scraper** (`scripts/scrape-motion-corpus.py`, offline dev tool): license-filtered discovery + download (Sketchfab Data API cc0/by, OpenGameArt, curated CC0 packs), per-asset provenance manifest (source URL, author, license, tags), `ATTRIBUTION.md` generation, validation via `qtmesh info --json` (skeleton + animations present). CC0/CC-BY only; Mixamo, LAFAN1, Bandai-Namco, AMASS and game rips explicitly excluded.
- **B — canonical normalization + template library v5**: retarget every scraped clip onto the 22-joint canonical skeleton (rig→canonical, the reverse of `applyMotionClip`; BVH path exists for CMU), active-window selection, keyword extraction from tags → a much larger clip library (hundreds of actions, several takes each). Immediate quality win with zero ML risk — the template path is the current quality bar.
[...]

PR #914, opened 2026-07-20
feat(anim): flip-facing toggle + world-facing metric (#837)

fernandotonon:
## Summary

Follow-up to the merged text-to-motion PR (#904), addressing the one remaining rough edge the user flagged: generated **walk/run clips face away from the default camera**.

**This is a camera convention, not an animation defect** — and I proved it rather than guessing. Clips face **+Z** (the Mixamo/GLTF default forward), which is *away* from a default viewport camera that looks toward +Z. Three prior attempts to "fix the facing" inside the retarget all made it worse; the disciplined measurement showed why:

| Clip | hip world-forward | faces |
|---|---|---|
| Model `generated_walk` | `(0.18, 0.14, 0.97)` | +Z |
| Rig's own `mixamo.com` clip | `(-0.10, 0.16, 0.98)` | +Z |
| Rumba | `(0.01, 0.01, 1.00)` | +Z |

All three face +Z — the model **matches the rig's native facing**. So the right fix is a user-facing toggle, not a retarget change.

## What this adds

- **`AnimationMerger::flipAnimationFacing(skel, anim)`** — turns a clip 180° about world +Y by rewriting **only the root (hips) track**. The whole skeleton turns rigidly through the hierarchy, so the pose (stride, arm swing, posture) is preserved **exactly**; only the body faces the other way. World pre-rotation `S = 180°@+Y` is folded into each root keyframe as `Lbind⁻¹·Wparent⁻¹·S·Wparent·Lbind·kf` (same bind-local conjugation as arm-space), and the keyframe translation is turned by `S` too so any root motion travels with the new facing. **Self-inverse** (two flips = identity) → a stateless toggle.
- **World-facing metric** — `qtmesh anim <file> --facing --animation <name>` plays the clip, samples the hip's world-forward (`fwd = left×up` from its bind frame) over 30 samples, and prints `+Z` / `−Z`. Read-only diagnostic.
- **Parity harness** — `qtmesh anim <file> --apply-canonical <clips.json>` (self-retarget round-trip) for numeric retarget validation.

## Surfaces (CLI / MCP / GUI parity)

- **CLI**: `qtmesh anim rigged.glb --flip-facing --animation generated_walk -o out.glb`; `--facing` metric.
- **MCP**: `flip_facing` tool (edits the master skeleton so `output_path` export includes it).
- **GUI**: per-row `⟳` button in the Inspector → Animations section (re-poses live even when the clip is paused).

## Verification

- **Numeric**: `--facing` on a flipped walk goes `(-0.00, 0.16, 0.99)` → `(-0.00, -0.16, -0.99)` — Z sign inverted, X/Y magnitudes unchanged ⇒ pose intact, body turned.
- **Render**: isometric sheet shows the body turned 180° with the upright stride and arm swing preserved.
- **Unit tests** (`AnimationMerger_test.cpp`): 180° turn reflects X/Z and preserves the inter-arm angle; self-inverse; no-op on missing anim / null skeleton. (Ogre-dependent fixture runs under Xvfb on Linux CI, same as the arm-space tests.)

## Summary by CodeRabbit

* **New Features**
  * Added a read-only `--facing` animation diagnostic that reports the character’s average world-forward direction in text or JSON.
  * Expanded text-to-motion action handling with more canonical action synonyms and improved template motion library guidance.
  * Added arm-space post-processing for more consistent retargeted animations.
* **Bug Fixes**
  * Improved motion-library quality checks with a stricter topple gate to reject unsuitable clips.
* **Documentation**
  * Updated guidance for motion-library retargeting, arm-space behavior, and third-party provenance/licensing and attribution.
* **Chores**
  * Streamlined CI SonarCloud analysis steps for the Linux unit-test workflow.


fernandotonon:
Addressed all three review comments in 568261e6:

- **Root-track rigidity (chatgpt-codex P2):** `flipAnimationFacing` now refuses clips it can't turn rigidly — the root track bone must be a plausible whole-body root (a true skeleton root, or the canonical hips role), and every other *animated* bone must descend from it. A partial/body-part clip whose highest track is a deep bone (e.g. `LeftArm`) is now rejected instead of flipping that subtree and corrupting the pose. Static armature ancestors above the hips stay fine (the common Mixamo layout). Added `FlipFacingRefusesPartialChildOnlyClip`.
- **Translate frame (coderabbit Major):** you're right that `S * t` is wrong, but the fix is **parent-space** `M = Wparent⁻¹·S·Wparent`, not the full bind-local `L`. `applyToNode` consumes the keyframe translation as a `TS_PARENT` offset *before* it rotates the node, so the translation lives in the parent frame — `L`'s extra `Lbind` conjugation only belongs on the rotation (which `applyToNode` post-multiplies). Verified end-to-end on the Rumba rig.
- **Missing breadcrumb (coderabbit Minor):** `anim --facing` now emits a Sentry breadcrumb like every sibling `cmdAnim` mode.

Issue #412, opened 2026-05-07
Epic: PS1 Runtime Geometry Extraction (experimental)


fernandotonon:
## Overview

Add a new **experimental** subsystem to QtMeshEditor — *PS1 Runtime Geometry Extraction* — that extracts geometry, textures, UVs, and vertex colors from PlayStation 1 games **at runtime** by embedding an emulator and intercepting GPU/GTE rendering commands, instead of trying to parse every proprietary PS1 game format individually.

The user loads an ISO they own, plays the game inside QtMeshEditor's emulator viewport, and clicks *Capture Frame / Capture Scene / Dump VRAM* to dump live mesh + texture data into the existing editor for cleanup and export (OBJ / glTF / Ogre `.mesh`).

This sits alongside the existing static parsers in `src/PS1/` (`PS1TMD`, `PS1TIM`, `PS1RSD`) — those keep working for known formats; this epic targets everything else.

## Motivation

- Most PS1 games ship custom, runtime-decompressed, per-engine mesh formats. Per-game format reverse-engineering doesn't scale.
- The PS1 GPU command set is **public, small, well-documented** (nocash psx-spx, ~30 GPU commands, 4 GTE matrix forms). A single GPU-side capture layer covers virtually every game.
- QtMeshEditor already has the downstream pieces: viewport, mesh editing, dedupe / weld / fix tools, animation merge, glTF/FBX/OBJ export, asset scan/lint. *Live ripping is the missing front-end.*
- This is a unique differentiator vs. Blender/Maya for retro / preservation / fan-remaster workflows.

## Scope

In scope:

- Embedding (or abstracting over) a permissively-licensed PS1 emulator core.
- A GPU-command interception layer that observes triangle / quad / textured / shaded primitives, plus GTE-derived world transforms and the camera/projection matrices.
- A VRAM observer that reconstructs texture pages + CLUT palettes into RGBA tiles.
- Capture & inspection UI (emulator viewport, capture buttons, VRAM viewer, geometry inspector, extracted-asset browser).
- Reconstruction heuristics: object grouping, deduplication, atlasing, coordinate-space normalization.
- Export to OBJ / glTF / Ogre `.mesh` reusing the existing `MeshImporterExporter` pipeline.
- Legality safeguards: user supplies their own ISO and BIOS; no asset bundling; clear UX warnings.

## Non-goals (this epic)

- Shipping a PS1 BIOS, ROMs, or commercial assets.
[...]

PR #917, opened 2026-07-21
feat(#412): merge same-object PS1 capture parts across frames

fernandotonon:
## What & why

PS1 scene captures ([#412](https://github.com/fernandotonon/QtMeshEditor/issues/412)) grouped prims by the **exact per-frame GTE matrix**, so a moving object — or any object seen while the camera moves — produced **one sparse part per frame**. Games NCLIP-cull back faces before GP0, so each frame only carries the camera-facing triangles. A dozen real objects in a scene reconstructed as **hundreds** of sparse fragments.

This adds a two-stage same-object merge to `MeshReconstructor`, **on by default** (the raw per-frame stream is unusable otherwise).

### Rigid stage — triangle union
Clusters tracked groups by **exact object-space vertex-set overlap**: the raw s16 GTE V registers are bit-identical across frames for rigid objects, so their per-frame parts chain (greedy union-find in capture order; ≥8 shared verts AND ≥25% of the group's set) and their triangles union into one part — **reconstructing faces no single frame showed**. Groups sharing a frame never merge, so simultaneous instances / hierarchy limbs (and their per-copy placement matrices) stay separate. A duplicate-triangle cull (canonical position+UV key; **colour excluded** because gouraud re-lights every frame) drops the once-per-frame repeats.

### Non-rigid stage — representative frame
Vertex-animated objects (CPU/GTE-warped menu text, breathing characters) defeat exact matching — their object-space verts differ every frame. Their per-frame clusters are chained by deformation-surviving signals (texture-key Jaccard ≥0.5, per-frame prim-count ratio ≥0.4, frame gap ≤3, and coarse 16-unit spatial-cell overlap ≥0.25 **or** stable draw-order rank ±1) and collapsed to the **best single frame** — a triangle union would superimpose every warp phase into ghost soup. Only short-span clusters (≤2 frames) may chain, so a genuinely rigid long-lived object (which always rigid-merges into one long-span cluster) can't be chained away and lose real geometry.

## Surfaces (all default-on)
- **GUI**: "Merge frames" toolbar checkbox (checked by default)
- **MCP**: `ps1rip_capture` `merge_objects` (default `true`)
- **CLI**: `--no-merge-objects` opt-out (`--merge-objects` kept as a back-compat no-op)
- Breadcrumb `ps1.rip.cleanup.merge_objects` + `MeshReconstructionStats` report `mergedGroups` / `nonRigidGroups` / `duplicateTrisDropped`.

Only in-core **tracked** captures are affected; RAM-scan captures have no object-space identity and are untouched.

## Testing
`MeshReconstructorMerge_test.cpp` (7 tests): rigid cross-frame union → full object, same-frame instance separation, static-repeat collapse, RAM-scan no-op, non-rigid representative-frame collapse, and the texture-gate separation of two alternating animated objects. Full PS1 suite (68 tests) green locally against `master` with `-DENABLE_PS1_RIP=ON`. Verified live on a real scene capture (menu / rotating props): part count drops from hundreds toward one-per-object.

## Limitations
For a warping object you get its best single captured frame, not a de-warped rest pose — that information isn't in the capture stream.

## Summary by CodeRabbit

* **New Features**
  * Same-object parts are now merged across captured frames by default, producing more complete reconstructed objects.
  * Added duplicate-triangle cleanup to reduce repeated geometry.
  * Added a “Merge frames” capture setting, enabled by default and saved between sessions.
  * Added CLI and capture-tool controls to enable or disable frame merging.
  * Added merge and cleanup statistics to capture diagnostics.

* **Documentation**
  * Documented merge behavior, controls, and reported statistics.


fernandotonon:
Addressed both review findings in 17885215:

**@codex (P2 — dup cull on non-tracked buckets):** Fixed. The duplicate-triangle cull is now gated to tracked groups (`key.first == 1`) only. Legacy matrixId groups (RAM-scan captures, and any stock capture where `mergeSameObjectGroups` did nothing) are never culled, so a coincident triangle drawn twice with different gouraud colours keeps both copies. Added a regression test (`LegacyCaptureCoincidentTrianglesSurviveMergeOn`) that fails without the gate.

**@coderabbitai (describe() deviation):** Fixed. Since `mergeSameObjectParts` now defaults to true, `describe()` emits `noMergeObjects` on the actual deviation (merge off) instead of `mergeObjects` when on — matching `isDefault()`'s semantics.

Full PS1 suite green locally (69 tests, incl. the 8 merge tests).

PR #942, opened 2026-08-09
feat(#506): Draco mesh compression on glTF export + Assimp 6.0.5

### LINKED ISSUE #506: feat(deps): rebuild Assimp with Draco support and enable KHR_draco_mesh_compression on glTF export
## Background

Phase 6 originally scoped a slice F for **Draco mesh compression** on glTF/glb export. Investigation during Phase 6 implementation revealed:

- Assimp 6.0.4 ships Draco source in `contrib/draco/` (already vendored in our build environment).
- The **prebuilt Assimp libraries** we link against (`/usr/local/lib/libassimp.dylib` on macOS, equivalents on Linux/Windows CI) were built with `ASSIMP_BUILD_DRACO=OFF`. Confirmed via:
  - `strings /usr/local/lib/libassimp.dylib | grep -i draco`
  - Output: `GLTF: Draco mesh compression not supported.` and `KHR_draco_mesh_compression` (the extension string + the not-supported warning the importer emits at runtime).

So the codebase recognizes the extension on import but **cannot produce** Draco-compressed glTF output. To ship slice F we need one of:

### Path A — rebuild Assimp with Draco (preferred)
1. Update the Assimp version pinned in `.github/workflows/deploy.yml` build matrix:
   - Linux: bump from the current `apt`-installed `libassimp-dev` 6.0.x to a source build with `-DASSIMP_BUILD_DRACO=ON -DASSIMP_BUILD_ASSIMP_TOOLS=OFF -DASSIMP_BUILD_TESTS=OFF`
   - macOS: rebuild via homebrew formula or vendored Assimp with `-DASSIMP_BUILD_DRACO=ON`
   - Windows: rebuild via the MinGW chain with Draco enabled
2. Add a `assimp_export_format=gltf2 ; aiExportFormatDesc::id=\"gltf2\"` invocation with the Draco-encoding hints (`ASSIMP_EXPORT_DRACO_COMPRESSION` or whatever the upstream API key turns out to be — see `assimp/include/assimp/Exporter.hpp`).
3. CI rebuild check across all three platforms.

### Path B — vendor Draco as a separate FetchContent dependency
Write a `MeshDracoEncoder` module that runs **after** Assimp's plain glTF exporter:
1. Read each `\"primitives\"` from the emitted glTF JSON
2. Pull the corresponding vertex/index buffers out of the binary chunk
3. Encode them with Draco
4. Rewrite the JSON to add `extensions: { KHR_draco_mesh_compression: { bufferView: N, attributes: {...} } }` per primitive
5. Append the Draco buffer to the binary chunk and update the buffer-view offsets

Estimated 600-1000 LoC. Carries its own copy of Draco (~30k LoC) via `FetchContent_Declare` — see existing pattern for `llama.cpp` / `stable-diffusion.cpp` / `googletest` in `CMakeLists.txt`. The Draco upstream CMake is well-behaved on macOS/Linux but historically painful on MinGW.

**Recommendation: Path A** — keep the rest of the codebase agnostic and let Assimp do what it already knows how to do. Path B is only worth it if the Assimp rebuild proves to be a CI nightmare across all three platforms.

## Acceptance criteria

- [ ] `qtmesh convert input.fbx -o output.glb --compress draco` produces a `.glb` whose `extensionsUsed` array contains `\"KHR_draco_mesh_compression\"` and whose vertex buffers are noticeably smaller (typical ratio 5-10× on triangulated meshes).
- [ ] Round-trip: `qtmesh convert output.glb -o roundtrip.fbx` produces a mesh with the same vertex/face counts as the original (lossless compression check).
- [ ] CLI exit codes: 0 on success, 1 if the build was compiled without Draco (with a clear error message pointing back to this issue).
- [ ] Unit test (`MeshDracoEncoder_test.cpp` or `ConvertWithDracoTest`): synthesize a Mixamo-style FBX, export with `--compress draco`, parse the resulting glTF JSON, assert the extension entry exists and that the binary chunk is smaller than the no-Draco baseline.
- [ ] Docs: CLI \`--help\` mentions `--compress draco`, website CmdSection for `convert` lists it, CLAUDE.md gains the dependency.
- [ ] CI: all three platforms (Linux / macOS / Windows MinGW) build cleanly with Draco enabled. Release artifacts include the Draco-aware Assimp binary.

## Context — what's already shipped

This issue completes the original Phase 6 scope. Adjacent slices already landed and pair well with Draco compression in a future \`qtmesh optimize\` pipeline:
- Slice C / C2 / C3 — vertex-cache optimization (lower ACMR before Draco compresses).
- Slice D — single-pass decimation (reduce tri count first).
- Slice E — texture atlas (consolidate texture bindings).

Slice F (Draco) was deliberately deferred from the immediate Phase 6 sequence because Path A is a CI-rebuild PR rather than an in-codebase feature PR.

fernandotonon:
Closes #506.

## What & why

Adds Draco mesh compression on glTF/glb export:

```
qtmesh convert model.fbx -o model.glb --compress draco
```

**Key finding that shaped the design:** Assimp's glTF2 *exporter* has **zero** Draco support — `ASSIMP_BUILD_DRACO=ON` only wires the Draco *decoder* into the *importer* (verified against the Assimp 6.0.5 source: `glTF2Exporter.cpp` has no draco references; the decoder lives behind `ASSIMP_ENABLE_DRACO` in `glTF2Asset.inl`). So rebuilding Assimp with Draco (the issue's "Path A") only enables *reading* compressed files, never writing them. The issue's **Path B — a standalone encoder module — is required**, and that's what this PR implements.

## Changes

- **`MeshDracoEncoder`** (`src/MeshDracoEncoder.{h,cpp}`) — Ogre-free, unit-tested post-processor. Parses the Assimp-written glTF (glb BIN chunk, or `.gltf` + data-URI/external `.bin`) with `QJsonDocument`; for each indexed-triangle primitive it Draco-encodes the geometry (POSITION/NORMAL/TEXCOORD/TANGENT, **float accessors only**; normalized-integer attrs and **JOINTS/WEIGHTS left uncompressed** to stay lossless), adds the `KHR_draco_mesh_compression` extension, strips the compressed accessors' `bufferView`, and **garbage-collects the orphaned geometry** so the file actually shrinks. Skins/morph targets/animations untouched. Re-emits glb or self-contained `.gltf`.
- **CLI** — `--compress draco` on `convert`, with up-front validation (unknown value / non-glTF output / build without Draco) and a size-reduction report.
- **Build** — `option(ENABLE_DRACO)` + `cmake/Draco.cmake` (discovers libdraco from the Assimp-with-Draco prefix or `-DDRACO_ROOT`); all Draco calls `#ifdef ENABLE_DRACO`-guarded, clean "rebuild with -DENABLE_DRACO" error otherwise.
- **CI** — `-DASSIMP_BUILD_DRACO=ON` on all 3 assimp builds; `-DENABLE_DRACO=ON` on the Linux release, Linux coverage/test, and macOS QtMeshEditor builds; caches busted. (Windows MinGW left off, matching ONNX/mocap.)
- **Assimp 6.0.4 → 6.0.5** (Ogre already at latest 14.5.2 on both local and CI — no change needed).
- Tests (`MeshDracoEncoder_test.cpp`), CLAUDE.md architecture entry, README, CLI help.

## Acceptance criteria

- [x] `qtmesh convert in.fbx -o out.glb --compress draco` produces compressed output
- [x] Round-trip maintains vertex/face counts
- [x] Proper exit codes + clear error messages
- [x] Unit tests validating compression + binary chunk size reduction
- [x] Docs updated (CLAUDE.md, README, CLI help)
- [x] Cross-platform CI builds with Draco support

## Verified locally (macOS arm64)

- `Rumba Dancing.fbx → .glb`: all 11 primitives compressed, **geometry 309 KB → 46 KB (85.1% smaller)**, whole file **761 KB → 449 KB**.
- Every primitive decodes back with **exact face + vertex counts** (checked with a standalone Draco decoder).
- Error paths (bad `--compress` value, non-glTF output) and the `.gltf` base64 path verified.
- 5/5 `MeshDracoEncoder` unit tests pass.

## Notes / follow-ups

- MCP + GUI parity for `--compress draco` is a natural follow-up (this PR is the CLI + core module).
- Draco encode is off on Windows MinGW for now (static-link, consistent with ONNX/mocap); import decode still works there via the Assimp build.

## Summary by CodeRabbit

* **New Features**
  * Added optional Draco compression for glTF, GLB, and VRM exports.
  * Added `--compress draco` support with compression statistics.
  * Added configurable compression and quantization settings.
  * Added validation and clear error reporting for unsupported formats, builds, and invalid files.

* **Documentation**
  * Added command-line examples and guidance for Draco-enabled builds.

* **Tests**
  * Added coverage for compression, decoding, file validation, size reduction, and unsupported configurations.


fernandotonon:
Thanks for the thorough review — all findings addressed in e70d09e6. Summary:

**Critical / P1**
- **Skinned/morph corruption (Codex L370, CodeRabbit L394):** compression is now **all-or-nothing per primitive**. Draco reorders/dedups points, so a primitive is compressed only if *every* per-vertex stream (POSITION/NORMAL/TEXCOORD/COLOR-float/TANGENT) can go into Draco; if it carries `JOINTS_n`/`WEIGHTS_n`, morph targets, or a non-FLOAT attribute, the **whole primitive is left uncompressed**. Skinned meshes are therefore not compressed yet (folding JOINTS/WEIGHTS into Draco losslessly is a tracked follow-up). New test `LeavesSkinnedPrimitiveUncompressed` locks this in.
- **Accessor counts after dedup (Codex L430):** compressed attribute accessors now get their `count` rewritten to the Draco decoded point count.
- **Index bounds (CodeRabbit L410)** and **compaction copy bounds (CodeRabbit L637):** both now validated before dereferencing/copying, returning a clean error instead of a heap over-read.
- **Draco not in caches (Codex L614):** every Assimp cache (producer + consumer) now includes `libdraco*` + `include/draco`.

**CI**
- Assimp's own `-DASSIMP_BUILD_DRACO` is broken on all three platforms (macOS `-Werror` on draco headers, Linux exports `assimp::draco` → an uninstalled `.so`, Windows/MinGW's `draco` install target doesn't exist). Switched to building the vendored `contrib/draco` as a **standalone static lib** and installing it beside Assimp — clean and identical on Linux + macOS. **Windows** stays off (CodeRabbit L211), consistent with ONNX/mocap; documented in the workflow + CLAUDE.md.

**Correctness / robustness**
- **GC conservative (CodeRabbit L619):** replaced the allow-list with a generic recursive scan of the whole JSON for `bufferView` refs, so EXT_meshopt / structural-metadata / vendor extensions are preserved. (Also fixed a bug this surfaced where the walk saw stale pre-strip accessors and kept all orphans alive.)
- **Atomic write (CodeRabbit L741):** temp file + verified byte count + flush + rename; a failed compression never truncates the already-written export.
- **`.glb2` (CodeRabbit L1882):** now accepted by `--compress` in both the CLI and `wantGlb`.
- **File-size report (CodeRabbit L1946):** output now reports the on-disk file delta *and* the geometry delta (a `.gltf` re-embeds base64 ~+33%, so geometry% alone could mislead).
- **File left on disk on failure (CodeRabbit L1936):** "0 eligible" is now a non-fatal warning (exit 0, valid uncompressed file kept); a hard failure explicitly notes the uncompressed file was written.
- **`#include <algorithm>` (CodeRabbit L97):** added.

Verified locally: static mesh 194 KB → 54 KB file (72% smaller, 0 orphan bytes); skinned Rumba warns + exits 0 with a valid uncompressed glb; 6/6 unit tests pass.

fernandotonon:
Pushed two follow-up fixes for the CI failures on the previous run:

1. **`unit-tests-linux` link error** (`36ab2507`'s parent `44e82ab9`): the `tests/` auxiliary test executables link a `qtmesh_test_common` static lib built from an explicit source list in `tests/CMakeLists.txt` that carried `CLIPipeline.cpp` (now calls MeshDracoEncoder) but not `MeshDracoEncoder.cpp` → `undefined reference to MeshDracoEncoder::isSupported()`. Added `MeshDracoEncoder.{cpp,h}` to that list + link `qtmesh_draco` into the common lib under `ENABLE_DRACO`. **The 6 Draco unit tests now pass in CI** (verified in the last run's log — the failure was only the aux-target link, the tests themselves ran green once linked).

2. **SonarCloud reliability gate** (`36ab2507`): Sonar flagged `MeshDracoEncoder.cpp:422` "uninitialized value" — `dracoAttrForSemantic()` only writes its out-params on success and two call sites ignored the return. Fixed by initializing at declaration + guarding the return, and by capturing the type/bits into `AttrPlan` once and reusing them (removing the second ignored-return lookup).

The earlier `build-windows` red was a transient Qt-installer flake (`aqtinstall: Specified path is bad: bin/cmake_automoc_parser.exe`), unrelated to this PR — this push re-triggers it. Linux + macOS app builds were already green.

PR #955, opened 2026-08-22
fix(mocap): stop body Neck drive from tipping FaceCap head

fernandotonon:
## Summary
- When **Head + Body** are both enabled, FaceCap owns the Head bone, but body retarget still aimed **Neck / Neck1** at the nose (webcam look-down). That pitch stacked with FaceCap and left the head tipped after calibration; head-only mode was fine because Neck stayed at bind.
- Skip the full face chain — `Neck | Neck1 | Head` — in live body retarget and body-clip bake whenever FaceCap drives the head.

## Test plan
- [ ] Enable Head only → calibrate → head stays level / tracks face as before
- [ ] Enable Head + Body → Recalibrate facing the camera → head no longer tips down after calibration
- [ ] Body limbs / torso still track; neck bones stay at bind while FaceCap drives Head
- [ ] Record a take with Head + Body and confirm the baked `_Body` clip does not include neck/head tracks that fight the face head clip

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented body tracking from overriding FaceCap head and neck positioning when head drive is enabled.
  * Fixed head tilting downward after calibration when combining body and facial tracking.
  * Ensured recorded body clips preserve the facial tracking head and neck pose.

### DIFF
--- a/src/Mocap/MocapController.cpp
+++ b/src/Mocap/MocapController.cpp
@@ -1051,7 +1051,8 @@ bool MocapController::beginPreviewWithLiveSource(
     d->savedAlwaysUpdateMainSkeleton = entity->getAlwaysUpdateMainSkeleton();
     entity->setAlwaysUpdateMainSkeleton(true);
     // Head-bone drive uses FaceCap (dense landmarks) even when Body is on —
-    // PoseIK's head role is coarse and fights the face solve.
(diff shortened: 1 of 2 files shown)


fernandotonon:
Addressed CodeRabbit: extracted `faceHeadChainSkipMask()` shared by live retarget and body bake.

Issue #549, opened 2026-05-17
Paint v2: Slice F — Projection / stencil painting + decals


fernandotonon:
Parent epic: #543
Depends on: Slice C (#546)

## Goal

Two image-driven paint modes:

- **Projection / stencil**: project the active viewport camera onto the mesh; brush stamps mask through a stencil image.
- **Decal**: place an image at a position/scale/rotation in the viewport, commit as a stamp into a new layer.

### Projection

- Stencil image alpha masks every stamp at the current screen position.
- **Camera-locked** mode: locks projection to active camera; rotating view doesn't re-project until user re-snaps.
- **Backface cull toggle** (Blender Texture Paint default).
- **Depth limit**: only paint within Z-distance of visible surface.
- **Project from photo**: load image, view-align through camera, commit as single stamp into a new layer.

### Decal

- New "Decal" tool. User picks an image; placeable rectangle handle appears in viewport (translate/rotate/scale).
- On commit, the rectangle's projection onto the mesh is rasterised into a new layer at proper UV resolution with soft edge falloff.

### Shared

- Both route through new `ProjectionPainter` (shares visible-surface ray-cast + UV-resolution with brush stamp pipeline).
- Both auto-create a new layer; don't stomp the active one.

## Acceptance Criteria

- [ ] Stencil projection paints only through stencil's visible region, masked to camera view.
- [ ] Backface cull + depth limit behave correctly on a sphere-with-hole fixture.
- [ ] Project-from-photo lands a recognisable single-stamp from an arbitrary photo.
- [ ] Decal placement handles drive easily (translate/rotate/scale; ESC cancels; Enter commits).
- [ ] Both modes create a new layer.
- [ ] Both modes undoable.
- [ ] Sentry breadcrumbs `paint.projection.*` / `paint.decal.*`.

## Effort

~10 days.

PR #956, opened 2026-08-23
feat(#549): Paint v2 Slice F — projection/stencil painting + decals

fernandotonon:
Implements issue #549 (Paint v2 Slice F). Builds on the Slice A–E texture painter.

## What

Two image-driven paint modes that project an image onto the mesh **through a camera**, rasterize it in UV0 space, and write into a paint layer. Both auto-create a **new layer** (never stomp the active one).

- **Projection / stencil** — brush through a stencil image, lock the projection to a camera pose, or project a photo straight onto the visible surface, with backface culling, per-texel occlusion, and a depth limit.
- **Decal** — a placeable image rectangle in the viewport (drag to move / rotate / scale, Enter commits, Esc cancels).

Full per-option user guide + architecture in **`docs/PAINT_V2_SLICE_F_DESIGN.md`**.

## Architecture
- **`src/ProjectionMath.h`** — header-only shared `projectToViewportUV` / `sampleImage`, extracted from `MultiViewTextureBaker` (its tests guard the no-op refactor).
- **`src/ProjectionPainter.{h,cpp}`** (pure-data) — the forked single-projection rasterizer: `project()` + `projectDab()`; per texel facing-cull → occlusion/depth-limit → sample → soft-edge → src-over composite.
- **`src/DecalSession.{h,cpp}`** (pure-data) — world-anchored oriented-quad state machine + orthographic `buildCommit`.
- **`TexturePaintController`** — projection state + stencil-brush hook, `snapProjectionCamera`/`projectFromPhoto`, and the decal session/overlay/commit. `MeshDepthRenderer::RenderResult` gained `depthNear/depthFar`.
- Viewport: `TransformOperator` decal mouse branch + `MainWindow` Enter/Esc.

**Occlusion** (the hard part): render a depth map from the camera (linear world distance via fog); per texel compare its **camera-axis** distance (not Euclidean — that self-occludes off-axis texels) against the reconstructed surface distance, biased above the 8-bit fog quantisation. The color `View` and the depth map's `viewProj` are kept strictly separate.

**Undo**: a photo/decal commit adds one `Generated` layer via `addFromBuffer` → one `PaintLayerOpCommand`. The stencil brush is an ordinary stroke.

## Acceptance criteria (#549)
- [x] Stencil projection paints only through the stencil's visible region, masked to the camera view
- [x] Backface cull + depth limit behave correctly (pure-data sphere-with-hole / two-plane occlusion + depth-limit tests)
- [x] Project-from-photo lands a single stamp from an arbitrary photo into a new layer
- [x] Decal placement handles (translate/rotate/scale; Esc cancels; Enter commits)
- [x] Both modes create a new layer
- [x] Both modes undoable (`PaintLayerOpCommand`)
- [x] Sentry breadcrumbs `paint.projection.*` / `paint.decal.*`

## Tests
- Pure-data (headless): `ProjectionPainter_test` (front projection, backface, stencil gating, dab, **sphere-with-hole occlusion**, depth-limit, self-projection no-acne) + `DecalSession_test` (transitions, hit-test zones, translate/rotate/scale, world↔rect-UV, ortho-commit corner→NDC, soft-edge).
- GL fixture (Xvfb on CI): projection-mode setters + graceful no-camera; decal begin/cancel plumbing.
- 85 paint/projection/decal tests pass; the existing 6 MultiViewTextureBaker tests still pass (refactor verified no-op).

## UI note
The **Projection** group in the Paint panel is collapsible (starts collapsed; a "•" marks an active mode) to keep the panel compact; the **Decal** tool has both a toolbar button and a panel row (Place / Commit / Cancel).

## Slices
F-A ProjectionPainter core · F-B occlusion + depth-limit · F-C projection UI + stencil + photo · F-D decal session + viewport handle · F-E toolbar + docs.

## Summary by CodeRabbit

* **New Features**
  * Added projection and stencil painting with camera locking, photo projection, backface culling, occlusion controls, and depth limits.
  * Added interactive decal placement with translation, rotation, scaling, commit, and cancellation workflows.
  * Added Projection and Decal controls to the Paint panel and toolbar.
  * Added separate paint layers for projections and contextual placement instructions.

* **Documentation**
  * Documented projection painting, stencil workflows, decal editing, controls, and undo behavior.

* **Tests**
  * Added coverage for projection, occlusion, decals, editing operations, and controller workflows.


fernandotonon:
Thanks both — all 10 findings addressed in ceecbc84.

**The real bug (flagged independently by both reviewers): `classifyDepth`.** Confirmed and fixed:
- The unconditional `biasWorld` occlusion test ran *before* the depth limit, so `return 2` needed `depthLimit < biasWorld` — the depth-limit branch was dead code and `texelsDepthCulled` never incremented.
- The existing test only passed because it set `biasWorld = 1e6f` to "isolate the depth limit" — that sentinel was masking the bug. It now uses a realistic `0.05` bias, and I verified it **fails against the old ordering** before restoring the fix.
- `useOcclusion` now gates the occlusion test, so Occlude-off + depth-limit no longer forces full occlusion culling. Added `DepthLimitOffAndOcclusionOffWritesThrough` to lock in the independence the design doc documents.

Also fixed:
- **Re-snap** — `snapProjectionCamera` went through `currentProjectionView()`, which returns the stored `m_lockedView` in locked mode, so it copied the stale pose onto itself. Split out `liveCameraView()`.
- **Decal normal** — now the inverse transpose, matching `MultiViewTextureBaker::fromEntity`'s `normalMat`.
- **Decal aspect ratio** — U extent derived from the image aspect. `scale()`/`worldToRectUv()` were already per-axis, so non-square rects need no other change. New `PlacePreservesImageAspectRatio` test.
- **`closeSession` notify** — emits `projectionChanged` (the NOTIFY for `decalSessionActive`/`decalState`); also resets `m_haveDecalDragPos` so a stale drag anchor can't leak into the next session.
- **Path privacy** — both breadcrumbs log `QFileInfo(...).fileName()` only.
- **Breadcrumbs** — added on the three projection setters and on the decal drag start/end (not in `mouseMoveEvent`).
- **Depth limit UI** — added the missing slider; the property was mirrored into `texPaintCol` with no writer, so the control was unreachable.

One extra: `ToolDecal` is now an explicit case in the per-dab tool switch (silences `-Wswitch`; behaviour unchanged, since a decal rasterizes once on commit rather than per dab).

Tests: 14/14 ProjectionPainter + DecalSession green. Note a pre-existing local-only segfault in `OgreWidgetTest::TearDownTestSuite` (`LightVisualizer` GL buffer teardown) reproduces identically with these changes stashed, so it is unrelated — CI runs tests on Linux under Xvfb and is green.

PR #960, opened 2026-08-24
feat(#549): WYSIWYG decal preview — show the artwork on the quad

fernandotonon:
Follow-up to Slice F (#956, merged). Requested in review of the decal tool: *"for the decal, or even the projection or the stencil, could we display a preview?"*

## Problem

The decal overlay drew its body as a flat translucent yellow tint, so you placed, rotated and scaled a **rectangle** without seeing what you were about to stamp. You only found out whether the placement was right after committing.

## Approach

The geometry was already correct — `refreshDecalOverlay()` draws two triangles at the exact rect corners and re-runs on every edit. It was just filling them with a solid colour. So this is **a material + UVs, not new plumbing**.

- `ensureDecalPreviewMaterial()` uploads the decal image to a manual GPU texture and builds an unlit, alpha-blended material that samples it — mirroring the existing `m_maskOverlayTex` pattern in this same file.
- The upload is keyed on `QImage::cacheKey()`, so drag/rotate/scale re-emit **geometry only**. Re-uploading per mouse-move would make transforms stutter.
- Falls back to the old flat tint if the texture/material can't be created, so the tool degrades rather than breaks.
- Torn down in `closeSession()` via `destroyDecalPreview()`, alongside the rest of the decal state.

### Depth-check asymmetry (deliberate)

The **artwork** gets `depthCheck = true`: the decal is pinned to a surface, so geometry in front of it should occlude the preview. The **outline and handles** keep `depthCheck = false` so they stay grabbable. Texturing the whole overlay uniformly would have made handles unclickable behind geometry.

### UV orientation was the trap

`buildCommit()` maps world→NDC and relies on `projectToViewportUV`'s flip (`v = 1 - (ndcY*0.5+0.5)`), so **`+tangentV` is image-UP**. The naive UV table renders the preview vertically **mirrored** against what actually gets baked — a bug that would look plausible until you compared before/after.

Rather than trust a reading of that convention, `CommitUvOrientationMatchesPreviewMapping` projects all four corners through the **real commit matrix** and asserts they land on exactly the UVs the preview uses. I verified it **fails when V is flipped**, so it genuinely pins the two conventions together instead of just documenting one — if either side changes, the test breaks loudly.

## Verification

- 15/15 `DecalSession` + `ProjectionPainter` tests green.
- Mutation-checked the new test (flip V → fails; restore → passes).
- Confirmed working in the running app by the requester.

## Scope

Decal only. The same question covered projection and stencil; both are feasible on this groundwork but are separate slices:

- **Stencil** is the natural next step — `m_stencilImage` is already loaded and there is an existing hover-ring overlay to hang it on. A cheaper first slice is a 2D panel thumbnail via the existing `data:image/png;base64` idiom.
- **Full projection preview** means running `ProjectionPainter::project` (with occlusion + depth-limit) into a scratch buffer and displaying it uncommitted, plus deciding when to re-run it — a proper slice, not an add-on.

## Summary by CodeRabbit

* **New Features**
  * Decal overlays now display the actual decal artwork in the viewport for a more accurate WYSIWYG preview.
  * Preview rendering supports decal transparency and refreshes when the artwork changes.
  * Dragging a decal across surfaces now repositions it while preserving its size and orientation.

* **Bug Fixes**
  * Corrected decal preview UV orientation so the viewport matches the final committed result.
  * Improved consistency between decal previews and committed images, including soft edges and transparent borders.
  * Improved handling of invalid or unavailable surface data during decal placement.

### DIFF
--- a/src/DecalSession.cpp
+++ b/src/DecalSession.cpp
@@ -88,6 +88,33 @@ void DecalSession::translate(const Ogre::Vector3& worldDelta)
     m_rect.center += worldDelta;
 }
 
+void DecalSession::reseat(const Ogre::Vector3& worldHit, const Ogre::Vector3& worldNormal)
+{
+    if (m_state != State::Editing) return;
+    Ogre::Vector3 n = worldNormal;
+    if (n.isZeroLength()) return;            // no usable normal — keep the old plane
+    n.normalise();
+
+    Ogre::Vector3 oldN = m_rect.normal;
+    if (oldN.isZeroLength()) oldN = Ogre::Vector3::UNIT_Z;
+    oldN.normalise();
+
+    // Rotate the existing tangent basis by the MINIMAL arc old->new normal, so the
+    // decal's size and in-plane orientation ride along instead of being rebuilt
+    // (a fresh place() would snap rotation back to camera-up and reset extents).
+    // getRotationTo handles the near-180 degree case by picking a perpendicular axis.
+    const Ogre::Quaternion q = oldN.getRotationTo(n);
+    if (!std::isfinite(q.w) || !std::isfinite(q.x)
+        || !std::isfinite(q.y) || !std::isfinite(q.z)) {
+        m_rect.center = worldHit;            // degenerate rotation: move only
+        return;
+    }
+    m_rect.tangentU = q * m_rect.tangentU;
+    m_rect.tangentV = q * m_rect.tangentV;
+    m_rect.normal   = n;
+    m_rect.center   = worldHit;
+}
+
 void DecalSession::rotate(float deltaRad)
 {
     if (m_state != State::Editing) return;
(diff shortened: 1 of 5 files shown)


fernandotonon:
All 3 findings addressed in 405d7718, plus a user-reported drag bug.

**1. Cache key was defeated (CodeRabbit, Major) — correct, and it contradicted my own commit message.** I keyed on `rgba.cacheKey()` taken *after* `convertToFormat()`, which returns a new QImage (new cacheKey) whenever the source is not already RGBA8888. So the cache missed every refresh and re-uploaded the whole texture **on every drag frame** — precisely the per-mouse-move upload I claimed it prevented. Now keys on the source `img.cacheKey()` before conversion, and the convert+feather work is gated behind that check too (it had been running unconditionally).

**2. Preview did not match the bake (Codex, P2) — confirmed.** `commitDecal()` feathers the outer 15% via `buildCommit(0.15f)` while the preview uploaded the raw image, so a decal with opaque edge pixels showed a hard border that softened only after commit. Extracted the feather loop into `DecalSession::featherSource()`, now shared by both paths, and replaced the hardcoded `0.15f` with `DecalSession::kDefaultSoftEdge` so they cannot drift apart again.

**3. Failed upload cached as success (CodeRabbit, Minor) — confirmed.** The `catch (...)` swallowed the error but still stamped `m_decalPreviewImageKey`, caching a blank quad that never retried. Now returns `{}` (flat-tint fallback) and leaves the key untouched so the next refresh retries.

## Also: the decal now follows the surface while dragging

Reported while testing — the decal kept the angle of its first placement. Body drag ray-cast against `decalPlaneHit` (the plane it was *first* placed on) and translated along it, so it slid across a frozen plane regardless of the surface underneath.

- New `DecalSession::reseat(hit, normal)` re-seats onto a new surface point + normal while **preserving half-extents and in-plane rotation**. That is what a naive fix gets wrong: re-calling `place()` snaps rotation back to camera-up and resets the size. It rotates the existing basis by the minimal arc between normals instead, with a finite check so a degenerate rotation moves the centre rather than emitting NaNs.
- `dragDecal` re-hits the real surface via `hitTestLocalPoint`, using the inverse-transpose normal transform so non-uniform scale cannot tilt the plane off the surface. Off-mesh cursor falls back to the old plane-slide.
- **Fixed a bug this would otherwise have introduced:** `dragDecal` opened with `if (!decalPlaneHit(...)) return;`. After re-seating onto a steep face the rect plane can go nearly edge-on to the camera, the plane ray misses, and the gesture would **die mid-drag**. The plane hit is now non-fatal for Body drags and still required for rotate/scale.

## Tests

18/18 green. Three new cases: preview-vs-commit pixel parity (plus an assertion that the feather actually *does* something, so both sides cannot "match" by both being unfeathered); reseat preserving size/rotation with an orthogonal basis; and reseat edge cases (~180° flip → finite basis, zero normal ignored, Idle no-op).

Mutation-checked the reseat tests — dropping the preserved size inside `reseat` fails both, restoring passes — so they guard the behaviour rather than merely executing it.

PR #962, opened 2026-08-25
feat(#550): Paint v2 Slice G — cavity / curvature / AO masks

### LINKED ISSUE #550: Paint v2: Slice G — Cavity / curvature / AO masks
Parent epic: #543
Depends on: Slice C (#546)

## Goal

Auto-generate per-mesh cavity, curvature, and AO maps; use them as brush colour sources or layer masks — the canonical way to add edge wear, crevice dirt, and weathering.

## Scope

- **Generators**:
  - **Cavity** — concave dark; per-vertex from normal angle to neighbours (via `HalfEdgeMesh`), then rasterised to UV space with seam dilation (reuse `VertexColorBaker` seam plumbing).
  - **Curvature** — signed cavity (concave -, convex +). Edge wear from positive curvature.
  - **AO** — short-ray hemispherical occlusion. CPU bake, few seconds on 5k-tri meshes.
- **Cache** in `<AppData>/paint/derived_maps/<mesh-hash>/`. Invalidate on topology change via EditableMesh revision counter.
- **Usage**:
  - **Brush colour source**: brush colour × derived map value at painted UV; combines with gradient/stamp colour.
  - **Layer mask**: a layer's mask initialised from a derived map; user paints on top to refine.
- **One-click recipes**:
  - "Add edge wear" — inverse-curvature-masked layer, default = light bare metal.
  - "Add crevice dirt" — cavity-masked layer, default = dark grime.
  - "Add AO darken" — multiply layer of AO onto BaseColor.
- "Recalculate derived maps" button (manual invalidate).

## Acceptance Criteria

- [ ] Cavity, curvature, AO compute correctly on a fixture with clear convex/concave regions.
- [ ] Maps cache + reload without recomputing.
- [ ] Each map can drive brush colour or initialise a layer mask.
- [ ] Each one-click recipe produces visibly correct results.
- [ ] Maps recompute on topology change.
- [ ] Sentry breadcrumb `paint.derived_map.*`.

## Effort

~7 days.

fernandotonon:
Closes #550. Auto-generated per-mesh **derived maps** that gate the brush or initialise a layer mask — the canonical edge-wear / crevice-dirt / weathering workflow. Design doc: `docs/PAINT_V2_SLICE_G_DESIGN.md`.

## The maps

| Map | Meaning | Use |
|---|---|---|
| **Cavity** | concave only (`0` flat/convex → `1` crevice) | crevice dirt |
| **Curvature** | signed (`0` convex ← `0.5` flat → `1` concave) | edge wear (inverted) |
| **AO** | occlusion (`0` open → `1` occluded) | weathering, contact shadow |

Cavity/curvature come from per-vertex concavity: the mean over the `HalfEdgeMesh` 1-ring of `dot(normal, normalize(neighbour - v))`. The mesh is welded across submeshes first, so a UV seam or material split does not read as a crease. Curvature pins near-flat values to exactly neutral — without that, tessellation noise on flat panels speckles into visible edge wear.

## Three places the issue did not match the code

**1. No `EditableMesh` revision counter exists.** The issue says to invalidate through one. Worse, `EditableMesh` exposes a public mutable `subMeshes()` accessor, so any counter could be bypassed without incrementing and would not be authoritative. Invalidation is instead by **content hash** — the SHA-1 already needed for the cache directory *is* the invalidation, with nothing to keep in sync. It covers positions/normals/UV/indices and deliberately excludes vertex colour and bone weights (they cannot change these maps, so including them would force needless rebakes).

**2. AO uses depth-map visibility, not CPU rays.** The issue specifies "short-ray hemispherical occlusion", but there is no BVH/kd-tree/octree anywhere in the repo and every existing ray query is a brute-force linear scan — a real ray AO meant adding an acceleration structure. This reuses the depth-map test `ProjectionPainter::OcclusionMap` already proves: render depth from 12 Fibonacci-lattice directions (a lat/long grid clusters at the poles and biases AO vertically) and count how many views can see each vertex. The visibility *maths* stays pure-data and headless-testable; only the rendering touches the Ogre scene.

**3. `VertexColorBaker` was not reused for rasterisation.** It is RGBA8-typed, so a scalar map would quantise to 8 bits and band visibly across a smooth AO gradient. I wrote a float rasteriser and copied its *discipline* — notably the explicit coverage vector, since a texel whose value equals the background is otherwise indistinguishable from an unwritten one.

## Non-obvious behaviours (all test-pinned)

- **Back-facing AO views are skipped, not counted as occluding.** Counting them darkens every vertex by ~half regardless of geometry.
- **No view facing the normal → `0` (unoccluded), not `1`.** The latter blacks out regions the view set happens not to cover.
- **A behind-eye-plane guard is explicit.** `projectToViewportUV`'s perspective `behind` (`w <= 0`) flag never fires under the orthographic-ish auto-framed depth views (`w` stays 1), so a point behind the camera got a *negative* axis distance that sailed through the `<= dMap + bias` test and read as visible. A test caught this.
- **Brush modulation scales ALPHA, not RGB.** Scaling RGB would drag paint toward black in cavities instead of hiding it there.
- **The two scalar fast paths are skipped while a map modulates.** Solid-colour and `GradientLinear` collapse the brush to one colour via the `paintBrush` overload that never calls `colorAt`; without this, enabling the mask would appear to do nothing for the most common brush setup.

## Layer masks + recipes

`PaintLayerStack::Layer::maskAlpha` and `ensureLayerMask()` already existed with the compositor honouring them, but had **no non-test caller** until now. "Mask active layer" fills it, sampling by UV rather than assuming 1:1 texels (a map may be baked at a different resolution than the paint buffer). Undoable via the existing `pushLayerOpUndo` path.

Three one-click recipes each add their own masked `Generated` layer as one undo step: **Edge wear** (inverted curvature / bare metal), **Crevice dirt** (cavity / dark grime), **AO darken** (AO / black / Multiply). A recipe temporarily switches kind to bake its own map and then **restores the user's picker selection**.

## Tests — 45/45

13 generator + 9 cache + 15 occlusion + 8 controller. Two worth calling out:

- **Cache tests are isolated** via `QStandardPaths::setTestModeEnabled` (the guard `BrushAssetLibrary_test` already uses). Without it they wrote to the real `<AppData>` *and read stale entries back* — a readiness assertion failed only because a `cavity.bin` from an earlier run was still on disk. That was the cache working correctly, not a bug; the polluted directory has been removed.
- **The brush-mask test was split after mutation testing.** The first version compared only `hasActiveSession()`, and a mutant returning factor 0 passed it. It now compares real pixels, and a second test bakes a real map and contrasts strength 1 vs strength 0 to actually exercise the wrapper.

## Verification status

Builds clean; `qmllint` reports zero errors on `PropertiesPanel.qml`. The QML group is verified **statically only** — the app launches and stays up, but its stdout/stderr are redirected internally so runtime QML warnings could not be captured from a shell here, and the bake/recipe buttons have **not** been exercised interactively.

Note the full test suite shows 14 pre-existing `AnimationProcessorChannelTest` failures plus a segfault under test-ordering pollution. I verified these reproduce **identically on master with these changes stashed** — unrelated to this slice.

## Known limits (documented)

AO bakes on the main thread (cached, so paid once per geometry); AO quality is bounded by the 12-view/256² budget; cavity/curvature detail is bounded by mesh density since they are per-vertex signals.

## Summary by CodeRabbit

- **New Features**
  - Added Cavity, Curvature, and Ambient Occlusion maps for texture painting.
  - Added controls to bake, recalculate, invert, adjust contrast and strength, and use maps as brush or layer masks.
  - Added Edge Wear, Crevice Dirt, and AO Darkening recipes.
  - Added caching to preserve generated maps and improve reuse.

- **Documentation**
  - Added design documentation covering derived-map workflows and behavior.

- **Bug Fixes**
  - Ensured derived-map design documents remain tracked.


fernandotonon:
Closing unmerged — the feature is not proven and is being parked.

Slice G built and tested cleanly (45/45, CI green including the Sonar gate), but it could never be demonstrated actually working on a real mesh. Testing on a Mixamo character surfaced a chain of defects:

1. **The picker early-returned on the default kind**, so clicking Cavity (the obvious first action) baked nothing.
2. **The raw geometric signal was too weak to see** — measured peak 0.56 with 78% of texels below 0.3, so the mask uniformly dimmed strokes instead of confining them. Auto-levelling fixed the distribution.
3. **The default paint target is Vertex**, and the modulation only exists on the texture path, so the mask was never consulted at all.
4. **An unexplained ~60x coverage shortfall** remained: the mesh's submesh 0 covers 41.3% of UV space but the bake reported 0.68%. The leading hypothesis (arithmetic only, unconfirmed) is that the map bakes at the paint buffer's resolution — which follows the *source texture* and can be 128px — while the stroke samples it across a 1024px buffer, making each map texel an 8x8 block that a small brush cannot resolve.

Diagnostics did clear the generator itself: a realistic bumpy grid at 41% UV coverage bakes 41.4% with a full 0..1 range, and seam-split and fully unwelded tri-soup meshes both bake correctly. So the pure-data core is sound; the failure is in how the controller feeds and samples it.

Two things worth carrying into any future attempt:

- **A documentation error to fix:** the design doc and commit messages claim `HalfEdgeMesh::buildFromEditableMesh` welds across submeshes so UV seams do not read as creases. It explicitly does **not** merge vertices across submeshes. That claim is wrong wherever it appears.
- **Scope question:** for AO specifically there is real overlap with simply painting the AO channel by hand. Cavity/curvature have no existing equivalent, so a future issue should probably narrow to those rather than re-litigating AO.

The branch `feat/550-paint-derived-maps` is left in place if any of this is worth salvaging later.

Issue #551, opened 2026-05-17
Paint v2: Slice H — Tablet pressure, brush presets, color palette


fernandotonon:
Parent epic: #543

## Goal

Three QoL features any serious paint app has:

- Tablet **pressure / tilt** modulating size + opacity + stamp rotation.
- **Brush preset library** for saving full brush configurations.
- **Color swatches / palettes** for curated reuse.

### Tablet

- `QTabletEvent` (cross-platform).
- Pressure curves: pressure→size (configurable, default linear), pressure→strength (configurable, default off).
- Tilt-X / tilt-Y → stamp rotation.
- Pen button mapping: lower = eyedropper hold, upper = erase hold.

### Brush presets

- Preset captures: tool, footprint mode, stamp image, radius, strength, falloff, colour source (solid / gradient with named ramp), spacing, scatter, jitter, channel, pressure curves.
- "Save current as preset" + named save.
- ~15 bundled presets: Soft round, Hard round, Pencil sketch, Spray paint, Foliage cluster, Edge wear, Scratched metal, Wet brush, Smudge soft, Smudge hard, Stencil hard, Stencil soft, Eraser soft, Eraser hard, Cavity dirt.
- Preset library panel: thumbnail grid, search, tags.
- Import / export presets as JSON.

### Palettes

- "Swatches" panel: 5×N grid of color slots.
- 6 bundled palettes (CC0): Material Design, Pantone classics, Skin tones, Foliage greens, Sky blues, Earth tones.
- **Recent colors** row (last 12 used).
- "Save palette from texture": extract representative colours from current diffuse.
- Storage in `<AppData>/paint/palettes/`.

## Acceptance Criteria

- [ ] Pressure modulates size + strength on a tablet.
- [ ] Tilt rotates stamp brushes correctly.
- [ ] At least 15 bundled brush presets apply cleanly.
- [ ] Save/load/import/export of custom presets works.
- [ ] At least 6 bundled palettes ship.
- [ ] Recent-colors updates as user paints.
- [ ] Sentry breadcrumbs `paint.tablet`, `paint.preset.apply`, `paint.palette.apply`.

## Effort

~7 days, plus 1–2 buffer for tablet quirks across platforms.

PR #970, opened 2026-09-01
feat(#551): Paint v2 Slice H — brush presets + colour palettes

fernandotonon:
Implements #551 **except tablet support** — see the scope note below. Design doc: `docs/PAINT_V2_SLICE_H_DESIGN.md`.

## Scope: tablet pressure/tilt is deliberately excluded

The issue bundles three features. Tablet pressure/tilt is **not implemented**, at the maintainer's direction: the project is desktop-only today, so pressure curves and tilt-driven stamp rotation could be written and unit-tested yet **never verified against real pen hardware**. Shipping an unverifiable headline feature is worse than deferring it; it should return as its own issue when there is a tablet to test on.

## Brush presets

A preset captures the whole brush: tool, footprint, stamp/tiling asset, radius / strength / falloff, edge shape, channel, stamp dynamics (spacing / scatter / size + opacity jitter / rotation), and the colour source (solid vs gradient + ramp name).

**15 bundled presets** — Soft/Hard Round, Pencil Sketch, Spray Paint, Foliage Cluster, Edge Wear, Scratched Metal, Wet Brush, Smudge Soft/Hard, Stencil Hard/Soft, Eraser Soft/Hard, Cavity Dirt — defined **in C++ rather than shipped as data files**, so they cannot go missing from an install (the `GradientRamp::bundledPresets` precedent). Custom presets are JSON in `<AppData>/paint/presets/`, with export/import to arbitrary paths.

Three behaviours that are easy to get wrong, all pinned by tests:

- **Applying a preset does NOT touch the paint colour.** A preset describes the brush, not what you paint *with*; restoring a saved colour on every click would silently discard the user's choice.
- **Apply order sets the stamp/tiling ASSET before the footprint TYPE** — the other order leaves a stamp brush briefly pointing at the previous preset's image.
- **Deleting a bundled preset is refused.** They are compiled in, so a delete could only remove a user override and the entry would appear to return on restart. The UI disables the button so the refusal is visible.

## Colour palettes

**6 bundled CC0 palettes** (Material Design, Pantone Classics, Skin Tones, Foliage Greens, Sky Blues, Earth Tones), a 5-column swatch grid (left-click = FG, right-click = BG, matching the existing FG/BG toolbar swatch), a 12-slot Recent row, and "Palette from texture".

- **Re-picking a colour promotes it** rather than appending a duplicate, so the recent ring stays 12 distinct colours. Only the **foreground** feeds it — the background is a rarely-changed secondary slot.
- **Extraction** uses 5-bit-per-channel colour-cube quantisation, averaging the real pixels per bucket so every swatch is a colour that actually occurs. Counting exact RGB values would return ten imperceptibly different shades. Fully transparent pixels are skipped, or a mostly-empty texture reports "transparent black" as its dominant colour.
- **`Swatch` carries no alpha** on purpose: a palette curates hues, and baking alpha in would override the user's brush alpha on every pick.

## Two panel defects fixed (reported from a screenshot)

- **Doubled separator above "Color:".** `syncShape` hides the "Edge:" row for Stamp/Tiling footprints but its separator stayed visible and collapsed against the next one — which is why the artefact only appeared in stamp mode. The separator now hides with its section.
- **Hint text rendering black.** It was `color: palette(mid)` — a mid-grey *border* role (its only other use in this file is borders), near-black on the dark theme. Now `palette(text)`, kept as a palette role rather than a literal white so it still follows the theme.

## Two documented departures from the issue text

- **Widgets, not QML.** CLAUDE.md says new UI should be QML, but these controls live inside the existing brush portal whose radius/strength/footprint controls are all Widgets; mixing toolkits in one panel buys nothing. Same reasoning as `StampLibraryDialog`, whose grid this follows.
- **Dropdown, not a thumbnail grid with search/tags.** Presets are parameter sets with no image to display, so a grid of identical tiles would carry no more information than the name. Search and tags are unjustified at 15 entries and become worth adding once a user library grows.

## Verification

**37/37** Slice H tests, plus **40/40** across the existing paint suites (BrushEngine / BrushFootprint / GradientRamp / PaintChannel / PaintLayer) to confirm no regression in the panel this touches. The UI was exercised interactively and confirmed working by the maintainer.

Two tests exist because of specific silent-failure modes:

- `BundledStampReferencesResolveToRealAssets` — a preset naming a missing stamp applies silently and leaves the previous footprint, presenting as "the preset did nothing". Mutation-checked: renaming "Charcoal" to "Charcol" fails it.
- `JsonFromOlderBuildKeepsDefaultsForMissingFields` — these files persist across versions, so a preset saved before a field existed must still load with the new field taking its struct default.

**A gotcha worth carrying forward:** `ApplyStampPresetSelectsItsStampAndFootprint` initially **passed against a mutant that skipped applying the stamp entirely**, because `m_activeStampName` is restored from QSettings and already held the expected value. The test now sets a different stamp first. Worth watching anywhere else that asserts against QSettings-backed controller state.

## Summary by CodeRabbit

* **New Features**
  * Added 15 built-in brush presets with support for saving, importing, exporting, and deleting custom presets.
  * Added six bundled colour palettes, swatch selection, recent colours, and palette creation from painted textures.
  * Updated the brush portal with preset controls, palette swatches, and foreground/background colour selection.

* **Bug Fixes**
  * Palette colour selection now preserves the brush’s existing alpha.

* **Documentation**
  * Added design documentation for brush presets, colour palettes, workflows, and limitations.

* **Tests**
  * Added comprehensive coverage for these features and their persistence.


fernandotonon:
All 6 findings addressed in 670ffdd7. Every one was real, and **two contradicted claims in my own design doc** — worth calling out rather than glossing.

**1. Swatches forced the brush opaque (Codex P2).** `QColor(r,g,b)` is alpha 255, so every palette pick silently discarded a partially transparent brush — the exact opposite of the "swatches carry no alpha so your brush alpha is untouched" rationale I wrote into the header and the design doc. Now copies the existing FG/BG alpha. Mutation-checked: removing the copy fails the new test.

**2. Recent ring missed colours set outside the palette (Codex P2).** Correct — the toolbar swatch and `pickBrushColorInteractive()` call `setVertexPaintColor` directly, so Recent only ever showed colours *re-picked from tiles*, never a newly chosen one. Now fed from `EditModeController::vertexPaintChanged`, with a no-op guard when the colour is already newest so the repeat notifications that signal emits cannot churn the ring.

**3 + 5. FG/BG ramp mode was neither saved nor restored (both reviewers).** Confirmed: `saveBrushPreset` always recorded `activeRampName()` and `applyBrushPreset` never called `setUseFgBgRamp`, so an FG/BG gradient preset came back using whichever *named* ramp was active. I added an explicit `useFgBgRamp` field to the struct and the JSON rather than treating an empty `rampName` as the sentinel — inferring it would make an older file that legitimately has no name silently become FG/BG.

**4. Non-atomic writes risked data loss (CodeRabbit, Major).** All three write paths truncated the destination first. Now `QSaveFile` (temp file + commit-by-rename) in all three, so a short or interrupted write preserves the previous file.

**6. A custom preset shadowing a bundled name could not be deleted (CodeRabbit).** This one was a trap of my own making: `allPresets()` lets custom override bundled, but `deleteBrushPreset` refused *any* bundled name — so saving a custom "Soft Round" would permanently hide the bundled one with no way back. Deletion now means "remove the custom file", gated by a new `canDeleteBrushPreset()` that the UI's Delete button shares.

**Tests:** 42/42 Slice H (4 new — alpha preservation, recent ring from the shared path, FG/BG round-trip, and deleting a shadowing override restoring the bundled preset), plus 40/40 across the existing paint suites to confirm no regression in the panel this touches.

Issue #818, opened 2026-07-06
Epic: AI v2 — working text-to-motion, non-human rigging & segmentation, next wave of local AI tools


fernandotonon:
Successor to #397 (closed) · Absorbs #793 (HY-Motion eval → Track A, Lane B) and #788 (meshseg retrain → Track B2) · Companion to the TripoSG work on `feat/triposg-backend` (#764 follow-up) and #805 (MV-Adapter)

# Epic: AI v2 — motion that works, non-human rigging & segmentation, and the next wave of local AI tools

## Overview

#397 shipped a broad local-AI toolset, but three gaps remain:

1. **Text-to-motion (#411) is the one feature that never reached production quality.** The template path is real mocap but only matches one clip per prompt keyword; the from-scratch CVAE is opt-in experimental and visibly below mocap quality. This epic ships a t2m that *works* — via composition first (guaranteed, all local, license-clean), generative second (opt-in).
2. **Segmentation and rigging are effectively humanoid-only.** MeshSeg's training labels derive from humanoid rigs, and the UniRig integration + the entire retarget stack assume the 22-joint humanoid canonical. Vegetation, vehicles, animals, and monsters need first-class support.
3. **The ecosystem moved.** Since #397's "no mature options" assessment: NVIDIA open-sourced Audio2Face-3D (MIT SDK + open weights), Microsoft released TRELLIS/TRELLIS.2 under MIT, and strong permissive small models exist for matting (BiRefNet, MIT), inpainting (LaMa, Apache-2.0), and depth (Depth-Anything-V2-**Small**, Apache-2.0). Several are drop-in upgrades to pipelines we already have.

Everything in this epic follows the project's hard rules: **local-only, no Python at runtime, permissive redistribution** (weights treated as derived works of training data — see `THIRD_PARTY_AI_MODELS.md` for the accept/reject history), **GUI + CLI + MCP parity**, models download-on-first-use from the HF `fernandotonon/QtMeshEditor-models` repo into `<AppData>/ai_models/<subdir>/`, Sentry `ai.assist.*` breadcrumbs, headless-CI-safe tests.

## What exists today (verified inventory — read before implementing anything)

| Feature | Files | Backend | Models | State |
|---|---|---|---|---|
| LOD / simplify (#398) | `src/MeshOptimizerLod.*` | meshoptimizer | — | mature |
| UV unwrap (#400) | `src/UvUnwrap.*` | xatlas | — | mature |
[...]

Issue #954, opened 2026-08-22
t2m retarget: roll baseline is reference-frame sensitive — library re-extraction changes arm bend-planes (fix: bind-relative transport, schema v5)


fernandotonon:
## Symptom
Re-extracting the motion library with the current importer produced clips whose ARM data is pitched 60-90° vs the (user-validated-good) July extractions: retargeted punches go overhead, arm swings render too wide, legs slightly crossed. Torso/legs extract consistently; only the arm chains are off.

## Evidence
- On `Animated Human by @Quaternius` (FBX, Blender export), the imported SKELETON REST has the arm bind directions along **−Z (horizontal)** while the legs point **−Y** — anatomically impossible for its arms-down idle bind (the DAE import of the same asset gives arms −Y, consistent).
- Old (July importer) vs new extraction of the same punch clip, canonical restDir:
  - relbow OLD `(0.50, 0.86, −0.12)` (forearm up = boxing guard ✓) vs NEW `(0.68, 0.11, −0.73)` (horizontal-back ✗) — 59°.
  - rshoulder 43°, lshoulder 70° — while abdomen differs only 4°.
- The extraction + retarget pipeline itself was verified transport-exact (in-process anatomy dumps, `QTMESH_T2M_DEBUG` on branch `feat/t2m-schema-v5-bind`): it faithfully reproduces whatever the import gives it.

## Suspected cause
The #936/#933 Blender-rig import changes (`BoneProcessor::bakeRootRotation` / node-chain orientation bake) correct the root/torso chain but leave (or double-apply on) the arm sub-chain for this rig class.

## Impact + mitigation (done)
The broken re-extraction had been uploaded to HF (`motion/motion-library-v2.json`) — **rolled back** to the July-extraction curated set on 2026-08-21. Library re-extraction is blocked until this is fixed.

## Repro
```
QTMESH_EXTRACT_DEBUG=1 qtmesh anim 'Animated Human.fbx' --dump-canonical /tmp/d.json --v2
# arm restDir vs leg restDir internally inconsistent; compare against the DAE variant
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

fernandotonon:
## Corrected diagnosis (2026-08-22)

[...]

PR #1022, opened 2026-09-14
feat(#1010): MotionComposer — multi-step prompts compose into one stitched clip

### LINKED ISSUE #1010: AI v2 A1: MotionComposer — prompt → action timeline → stitched mocap
Part of #818 (Track A1) — **the epic's headline feature.** Status: **ABSENT.**

## Problem
`MotionLibrary::matchPrompt` maps a prompt to exactly ONE action and returns ONE clip index. So "walk forward, sit down, wave twice, then stand up" plays a single clip. This is diagnosis #1 in the epic body and it is still verbatim true.

## Design (from #818 A1)
1. **Prompt parser** — `MotionComposer::parse(prompt) → MotionScript`. Primary: local LLM via `LLMManager` with a GBNF grammar so output is always valid JSON:
   ```json
   {"steps":[{"action":"walk","duration_s":2.0,"repeat":1,"direction":"left"},
             {"action":"sit"},{"action":"wave","repeat":2},{"action":"stand"}]}
   ```
   **Deterministic fallback with no LLM installed**: rule-based splitter on connectives ("then", "and", "after", "N times", digits) feeding the existing synonym map. The feature must not hard-require a GGUF download.
2. **Timeline compiler** — per step pick a take (reuse the quality-weighted selection), cut to duration, apply `repeat` by loop-splicing at the calmest frame, `direction` by yawing root motion.
3. **Transition stitcher** — between steps find the best cut pair (last 0.5 s x first 0.5 s, cost = summed quaternion geodesic over the 22 joints, feet weighted 2x), then bridge with **RMIB in-betweening** (`MotionInbetween`, 8-frame window — what it was trained for) rather than a crossfade. Accumulate hip trajectory across steps so the character continues from where it is (no teleports).
4. **Output** — one `MotionClip` in the existing world-space canonical format -> existing `applyMotionClip` -> existing playback. No new renderer/rig code.
5. **Surfaces** — GUI "Generate from text" composes transparently on multi-step prompts; CLI `qtmesh anim --generate` unchanged; MCP `generate_motion` gains an optional `script` param taking the JSON directly. Breadcrumb `ai.assist.motion.composed`.

## Files
New `src/MotionComposer.{h,cpp}` (Ogre-free, unit-testable like `MotionLibrary`); touch `MotionLibrary` (expose clip metadata + take enumeration), `CLIPipeline`/`MCPServer` wiring. `AnimationMerger` likely unchanged.

## Acceptance
The four-step example yields one fluid clip with no teleport/pop at seams, on a Mixamo-style rig AND an auto-rigged mesh; works offline with no LLM (degraded parsing, same stitching). Tests: parser goldens on the deterministic path (no LLM in CI); seam continuity (max joint angular velocity at seams <= 1.5x intra-clip max; hip position continuous).

fernandotonon:
Implements **#1010** (epic #818 Track A1) — the epic's headline feature.

## The problem

`MotionLibrary::matchPrompt` maps a whole prompt to exactly ONE action and returns ONE clip. So the epic's own acceptance prompt —

> "walk forward, sit down, wave twice, then stand up"

— played a single walk clip. This was diagnosis #1 in the epic body and was still verbatim true.

## What this does

New `src/MotionComposer.{h,cpp}` (Ogre-free, unit-tested like `MotionLibrary`), in three independently testable stages:

1. **parse** — deterministic connective splitter ("then", "and then", "after that", ",", ";", "finally") reading per-step repeat ("twice", "3 times") and duration ("for 2 seconds"). Runs with **no LLM installed**, which the epic explicitly requires. An LLM or agent can instead supply the same timeline as JSON via `parseJson`.
2. **compile** — each step picks a take through the library's own quality²/posture weighting, cut to duration and spliced for repeats.
3. **stitch** — consecutive takes join at their best-matching pose pair (quaternion geodesic over the 22 canonical joints, feet weighted 2×), then crossfade over a short window so junctions don't pop.

Output is shaped exactly like a `MotionLibrary::Clip` payload, so it feeds the existing `applyMotionClip` retarget with **no new rig or renderer code**.

`MotionLibrary` gains `resolveAction` / `takesForAction` / `pickTake`, and `matchPrompt` is refactored onto those same primitives — so the composer reuses the shipped synonym table and take-selection rule rather than growing a second vocabulary that could drift. Behaviour of `matchPrompt` is unchanged.

## Surfaces (full parity)

- **CLI** — `qtmesh anim rigged.glb --generate "walk then sit then wave twice"` reports the step list and emits one stitched clip.
- **MCP** — `generate_motion` composes multi-step prompts and gains the epic's `script` parameter for supplying a timeline directly. Responses report `composed_steps` and `unresolved`.
- **GUI** — the existing "Generate from text" box composes transparently; the status line says "composed from N steps".

**Single-action prompts are bit-identical to before** — `selectForPrompt` delegates straight to `matchPrompt`, keeping the V1 finger side-channel (deliberately dropped across a composition, where splicing two takes' curls at a seam would apply one take's finger timing to the other's body).

## Verification

End-to-end on `Goblin_rigged.glb`:

| input | result |
|---|---|
| `"walk then sit then wave twice"` (CLI) | one 286-frame / 9.5 s clip, 17 canonical joints, foot-pinning applied |
| script timeline (MCP) | `generated_walk_sit_wave`, 282 frames, `composed_steps: [walk, sit, wave]` |
| `"run then juggle chainsaws then jump"` | `run_jump` + `unresolved: ["juggle chainsaws"]` — reported, never silently played |
| `"walk"` | `generated_walk`, unchanged, no `composed_steps` |

Rendered frame-by-frame: walk → crouch → raised-arm wave, in order.

**20 unit tests**, validated by mutation testing — ignoring `repeat`, returning a trivial cut, dropping the seam blend, and ignoring `duration` each fail a test. That process caught two fixture bugs worth noting, both of which made assertions vacuous rather than wrong:
- overlapping clip angle bands let `bestCut` find a perfect seam, so the seam step was exactly `0.000000` and any blend assertion passed with blending removed;
- identical filler poses made the cut-search assertion ambiguous (several pairs tied at distance 0).

The 96 pre-existing `AnimationControlController` failures in a headless local run are unrelated — identical with this branch stashed (they need the GL context CI provides).

Closes #1010.

## Summary by CodeRabbit

- **New Features**
  - Added multi-step text-to-motion generation, combining actions such as walking, sitting, and waving into one continuous clip.
  - Added support for scripted motion timelines using JSON input.
  - Added repeat counts and per-step duration controls.
  - Responses now report composed steps and unresolved actions.
  - Improved error messages for unknown or unavailable actions.

- **Bug Fixes**
  - Duration retiming now works consistently for both single-action and composed animations.


fernandotonon:
Thanks — worked through both findings.

**1. Vertical descent on composed clips — valid, fixed in bd323d7b.**

Confirmed the bug: `isVerticalDescentAction` matches exact canonical labels, so a composed name like `walk_sit_wave` matched nothing and the sit lost its crouch.

I did *not* take the suggested "evaluate the name's individual components" approach, because it is wrong in the other direction — it would apply descent to the whole clip and sink the walk segment too. `rootY` is already per-FRAME and the composition carries each step's own values (locomotion frames sit at ~0), so the correct fix is to open the per-clip gate when ANY step is a descent action and let the per-frame data do the rest.

`Selection` now reports `verticalDescent`; CLI, MCP and GUI all read it instead of re-deriving from the clip name, and the single-clip/`--variant` paths set it from the library exactly as before.

Verified end-to-end: `"walk then sit"` reports `(grounded to feet)`, `"walk then run"` does not. Two unit tests pin both directions.

**2. GUI not passing `refRoll` — real, but pre-existing and out of scope here.**

`AnimationControlController.cpp` has **zero** occurrences of `refRoll`, on master as well as on this branch (`git show master:src/AnimationControlController.cpp | grep -c refRoll` → 0). So the GUI never passed bind-anchored roll; this PR did not regress it.

I have left it alone deliberately — adding a #954 argument the GUI has never passed is a behaviour change to every GUI-generated clip, which belongs in its own PR with its own before/after render check, not smuggled into a composition PR. Worth filing separately.

PR #1024, opened 2026-09-14
feat(#1009): motion library loop metadata + loop-aware repeats

### LINKED ISSUE #1009: AI v2 A2: motion library loop metadata (loopable, loop_start/end, in_place, tags)
Part of #818 (Track A2). Status: **PARTIAL — coverage done, loop metadata missing.**

## Done
The corpus/coverage half shipped under epic #837: `scripts/build-motion-library-v5.py` (and the later v6 builder), the CC0/CC-BY scraper (#838), and per-clip quality scoring + weighted take selection (#855). Schema reached `qtmesh-motion-library-v4`.

## Remaining
Per-clip **loop metadata**, which A1 (#TBD MotionComposer) needs to repeat an action cleanly ("wave twice"):

- `loopable: bool`, `loop_start` / `loop_end` frames
- `in_place: bool`, `travel_axis`
- tags (e.g. `upper_body_only`, which later enables layering "wave while walking")

`MotionLibrary::Clip` currently has no loop fields; the builders only mention loop seams in comments (heuristics, nothing emitted).

## Acceptance
- Library JSON stays within budget; loader remains backward-compatible (old editors keep working).
- Loop points verified by a script: first/last-frame canonical pose distance ≤ threshold.

fernandotonon:
Implements **#1009** (epic #818 Track A2's remaining half).

**Stacked on #1022** — review/merge that first; this branch contains it.

## What was missing

A2's coverage half shipped under epic #837 (the v5/v6 corpus builders, quality scoring). The **loop metadata** half never did: `MotionLibrary::Clip` had no loop fields, and the builders only mention loop seams in comments.

## What this adds

`Clip` gains `loopable` / `loopStart` / `loopEnd`, read from the library JSON when a builder supplies them and otherwise **derived at parse time** — so the 122-clip library already on users' disks gains loop points with no regeneration and no redownload.

## The detection is not "smallest seam distance"

That was my first implementation and measurement showed it is wrong. Most clips open *and* close near a rest pose, and a one-shot that HOLDS its final pose has dozens of near-identical late frame pairs. Measured on the shipped library:

| clip | chosen range | span | energy in range | seam |
|---|---|---|---|---|
| death | 98..106 | 8% | 0.16 | 0.46 ✗ "perfect loop" |
| run | 0..19 | 100% | 74.9 | 0.13 ✓ |

So seam quality is a **gate**, and coverage is what gets maximised among ranges that pass it. Blending the two as a single score fails: a flawless-seam tail (0.133) beats a true cycle with a looser seam (0.111). `loopable` then requires **both** a tight seam and a range spanning most of the clip's motion — a held end pose satisfies the seam alone.

## Thresholds are calibrated, not guessed

Measured across all 122 shipped clips; the two groups separate by orders of magnitude:

- **loopable** — every `*loop`-named clip ≤ 0.18; wave 0.0006, walk 0.0008, jump 0.006, idle 0.016, run 0.13, attack 0.46
- **one-shot** — death 18.7 (median), pickup 36.1, landright 4.1

The 1.0 gate sits in the empty band between them.

## Composer uses it

`"wave twice"` now cycles the wave itself instead of replaying its lead-in and settle each time; non-loopable one-shots repeat whole, the only sensible reading of "do it twice".

That exposed a real bug in #1022, fixed here: `selectForPrompt` short-circuited **any** single step to `matchPrompt`, which knows nothing about `repeat`/`duration` — so `"wave twice"` silently played once (73 frames for both, on the real library). Only a step with neither may take that shortcut. Now 73 / 146 / 219 frames for ×1/×2/×3.

## Verification

45 unit tests. Mutation-tested: ignoring the coverage condition, marking everything loopable, and a sign-sensitive pose distance each fail a test. The first needed a fixture carrying a genuine still tail — my initial synthetic clip was too clean to reproduce the trap the real data showed.

Closes #1009.

## Summary by CodeRabbit

- **New Features**
  - Motion clips can automatically detect suitable loop ranges when loop metadata is unavailable.
  - Repeated loopable motions replay only the designated loop section for smoother repetition.
  - Finger movement data is preserved when composing or repeating a single motion take.

- **Bug Fixes**
  - Improved handling of repeats and duration limits, including cuts before a loop begins.
  - Prevented finger data from carrying into compositions assembled from multiple motion takes.
  - Explicit motion scripts now take precedence over prompt text, including empty or unrelated prompts.
  - Clips with incomplete loop metadata are no longer incorrectly treated as loopable.


fernandotonon:
Both findings were real regressions from this branch. Fixed in 9a614308, each reproduced as a failing test first.

**1. Finger side-channel lost on single-step repeats — valid.**

Making `"wave twice"` compose (so the repeat is actually honoured) routed it away from the single-clip branch, which was the only place setting `sel.fingers`. So on a V1 library any single-step prompt carrying a repeat or duration silently lost fingers.

Fingers now travel with their own take's frames through the compiler and are emitted when the whole composition came from **one** take (`Selection::singleTake`). Across several takes they remain dropped deliberately — the seam blend reflows the body while the finger channel cannot follow, so one take's curls landing on another take's pose is worse than shipping none. I added a test for that direction too, because it is the more dangerous one.

**2. Empty loop pass when the duration cuts before `loopStart` — valid.**

Confirmed: with `repeat: 3` and a duration short enough that `keep <= loopStart`, every pass after the first had `from > to` and contributed nothing — 3 frames for `repeat=3`, identical to `repeat=1`. The loop slice now requires `loopStart < keep`, with a `from > to` guard as belt-and-braces.

Worth noting the fixture had to be built explicitly (`loop_start`/`loop_end` in the JSON): a synthetic clip is not loopable with a late start by accident, so my first attempt at this test passed against the bug.

**Mutation-tested.** Removing the `loopStart` guard, dropping the finger carry, and forcing `singleTake = true` each fail a test. The third initially survived — the multi-take test above was written specifically to kill it.

**Verified on the real 122-clip library:** 73 / 146 / 219 frames for wave ×1/×2/×3, and `"walk then sit then wave"` still composes 3 steps.

Note this PR is stacked on #1022 (now fully green), so CI here is skipped until that merges and this retargets to master.

fernandotonon:
## Conflicts resolved

#1022 was **squash-merged**, so master gained one commit (`f912b186`) containing changes my branch still carried as five separate commits. Git saw the same content twice, which is where the wide conflicts came from — not a genuine content clash.

Rather than rebase across commits already in master, I rebuilt the branch from `origin/master` and applied the net diff. The result is **one commit, six files, zero conflict markers in a trial merge** (`git merge-tree` → 0). PR now reports `MERGEABLE`.

Nothing was lost in the rebuild — verified both the composer features that came from #1022 (`selectForPrompt`, `bestCut`, `parseJson`, `verticalDescent`, `singleTake`) and the new loop work (`findLoopRange`, `loopable`, `energyCoverage`, calibrated thresholds) are present, with 48 tests passing.

## Review findings

Both P2 findings were already addressed in `9a614308` before the rebuild, and I re-verified they survived it:

**1. Finger data lost on modified single-clip selections** — correct, and it was a regression from this branch. Fingers now travel with their own take's frames through the compiler and are emitted when the whole composition came from **one** take (`Selection::singleTake`). Across several takes they stay dropped deliberately: the seam blend reflows the body while the V1 finger channel cannot follow, so one take's curls landing on another take's pose is worse than shipping none.

**2. Repeats lost when the duration ends before the loop range** — also correct. With `keep <= loopStart` every pass after the first had `from > to` and contributed nothing (3 frames for `repeat=3`, identical to `repeat=1`). Loop slicing is now disabled when `loopStart >= keep`, with a `from > to` guard as belt-and-braces.

Both reproduced as failing tests first. Finding 2 needed an explicit `loop_start`/`loop_end` fixture — a synthetic clip is not loopable with a late start by accident, so my first attempt at that test passed against the bug.

**Mutation-tested on the rebuilt branch**, all three killed:

| mutation | test that catches it |
|---|---|
| remove the `loopStart < keep` guard | `RepeatIsHonouredWhenDurationCutsBeforeTheLoopStart` |
| drop the finger carry | `SingleTakeRepeatKeepsFingerSideChannel` |
| force `singleTake = true` | `MultiTakeCompositionDropsFingers` |

The third needed a test written specifically for it — it survived the first round, and it is the dangerous direction (a multi-take clip emitting mismatched fingers).

**Verified on the real 122-clip library:** 73 / 146 / 219 frames for wave ×1/×2/×3, and `"walk then sit"` composes 2 steps and still reports `(grounded to feet)`.

CI is running properly now — it was skipped while the base was a branch.

fernandotonon:
Both findings from this round were valid and are fixed in `0fdee9b2`.

**1. Explicit script ignored (major) — confirmed and fixed.**

You were right about the mechanism: the shortcut re-runs `matchPrompt(prompt)`, which searches only the prompt text. Reproduced both failure modes as a test before fixing:

- `{"steps":[{"action":"wave"}]}` with **no prompt** → `no motion matched ""` (hard failure)
- the same script with prompt `"walk"` → silently played **walk**

I took your suggested gate (`scriptJson.isEmpty() && …`), and extended it to the `steps.empty()` arm too — otherwise a script whose steps all fail to resolve would fall through to the prompt and play something the caller never asked for. That case now reports an error with the unresolved action listed.

Verified live over MCP against `Goblin_rigged.glb`: script-only → `wave` / 73 frames (was an error); `"walk"` prompt + wave script → `wave`; plain `"walk"` prompt → `walk` / 61 frames, unchanged.

**2. Bounds-only metadata treated as loopable (minor) — confirmed and fixed**, though I landed on a slightly different shape than proposed.

I first did exactly what you suggested — require all three fields — and then **removed that guard again**. With the flag defaulting to `false`, requiring it to be present is behaviourally identical: mutation testing showed `loopable-not-required` and `loopable-defaults-true` were both **equivalent mutants**, i.e. no test could distinguish the two guards because each masked the other.

So the fix is the `false` default alone, which is pinned by a test. One guard that a test can hold is better than two that make each other untestable.

**Tests.** Four added. The bounds-only case cannot pin the explicit branch on its own — it asserts the same `false` whether the branch is taken or not — so I added both directions: an explicit `loopable:true` must override a derivation that would reject, and an explicit `false` must veto one that would accept.

Mutation-tested: ungating the shortcut, defaulting `loopable` to true, and ignoring the explicit flag each fail a test.

53 tests passing. All four platform builds were already green on the previous commit; this pushes a new run.

Issue #1025, opened 2026-09-14
UniRig: hosted encoder.onnx outputs 100% NaN latents — --algo unirig has never run the ML model


fernandotonon:
The hosted UniRig encoder export produces **100% NaN latents** on every input. `--algo unirig` has therefore never actually run the ML model — every call silently falls back to the Pinocchio template rig.

Found while starting #1013 (non-human categories). It blocks that issue, and any other UniRig work.

## Evidence

Encoder inputs are clean; the encoder output is structurally correct and entirely non-finite:

```
[unirig] enc.pts   n=98304   nonfinite=0  range=[-0.9997, 0.9994]
[unirig] enc.nrm   n=98304   nonfinite=0  range=[-1.0000, 1.0000]
[unirig] encOut[0] 'latents' shape=1x1024x1024 n=1048576 nonfinite=1048576 first4=nan nan nan nan
[unirig] dead state: step=2 validTokens=260 haveAnyJoint=0 top5=2(nan),4(nan),0(nan),1(nan),3(nan)
```

The decode then dies at step 2 — not because the FSM is wrong (260 tokens are valid) but because every logit is NaN, so no token can be selected. That surfaces as the existing message *"UniRig: constrained decode reached a dead state."*

## What it is NOT

Each of these was tested and ruled out:

- **Not input preprocessing** — points normalise to [-1,1], normals are unit, zero non-finite values in either tensor.
- **Not CoreML** — `QTMESH_ONNX_FORCE_CPU=1` gives byte-identical all-NaN output, so it is not an EP numerical issue.
- **Not mesh-specific** — reproduced on a humanoid (`jana.obj`), a car (`buick_riviera*.glb`), i.e. it fails on the class the checkpoint was trained for, not just non-human input.
- **Not a truncated download** — local files match the hosted sizes exactly: encoder 216,674,173 / decoder 1,223,378,677 / embed 1,093,821 bytes.
- **Not a wiring/shape bug** — the output shape `1x1024x1024` is exactly the expected latent prefix.

So the fault is in the exported graph itself (`scripts/export-unirig-onnx.py`), not in `UniRigPredictor`.

## Impact

[...]

PR #1027, opened 2026-09-14
feat(#1018): photo depth estimation via Depth-Anything-V2-Small (exported, parity-verified, hosted)

### LINKED ISSUE #1018: AI v2 C4: Depth-Anything-V2-Small photo depth (Small only — Base/Large are NC)
Part of #818 (Track C4). One of the epic's three named quick wins. Status: **ABSENT.**

Depth-Anything-V2-**Small** only (Apache-2.0). **Base/Large are CC-BY-NC — do not ship them.** ~25 M params, ONNX-friendly, CPU-fast.

Consumers:
1. Depth-condition the SD ControlNet texture path from a **reference photo** rather than only mesh-rendered depth (`SDWorker::generateTextureControlled` already takes a control image — this just produces better ones).
2. Height/displacement from a photo for the PBR pipeline (feeds `NormalMapGenerator` + parallax).
3. Later: depth prior for single-image reconstruction QA.

New `src/PhotoDepth.{h,cpp}`, `ai_models/depth/da2_small.onnx` (~50 MB), CLI `qtmesh material photo-depth in.jpg -o depth.png`, MCP tool.

NB `src/MeshDepthRenderer.*` is the mesh-depth RENDERER for ControlNet — unrelated to photo depth ESTIMATION; do not conflate them.

fernandotonon:
Implements **#1018** (epic #818 Track C4) — one of the epic's named quick wins.

The model is **exported, parity-verified and hosted**, so this works on first use rather than shipping plumbing that waits for weights.

## What ships

New `src/PhotoDepth.{h,cpp}` (pure data, no Ogre/GL — the `PbrMapSynth` shape) plus `scripts/export-depth-anything-onnx.py`.

```console
qtmesh material --photo-depth photo.png [-o depth.png] [--depth-letterbox]
```

Output is 8-bit grayscale, **near = bright** — the same convention `MeshDepthRenderer` emits, so a photo and a rendered mesh can feed `SDWorker::generateTextureControlled` interchangeably. That was the point of the issue: depth-condition the SD texture path from a *reference photo* rather than only mesh-rendered depth.

## Licence: Small only, enforced in code

Small is **Apache-2.0**. Base and Large are **cc-by-nc-4.0** and can never ship — the same bar that rejected SF3D and Hunyuan3D. Rather than leave that to reviewer vigilance, the export script queries the HF model API and **hard-fails before downloading any weights**. Verified by running it against `Base-hf`:

```
REFUSING to export depth-anything/Depth-Anything-V2-Base-hf: license
'cc-by-nc-4.0' is not in the permissive allow-list ['apache-2.0'].
```

No file was written.

## Two export defects found and fixed

Both are the kind that ship silently:

1. **The obvious dynamic-axis export freezes the output size.** A 462×462 input still produced a 518×518 map — `int()` casts in the HF head bake the interpolate target into a TorchScript constant. The graph *advertises* dynamic H/W, so nothing downstream would have noticed.
2. Re-interpolating to the input size fixed the shape and then exposed genuine **accuracy drift** away from the traced resolution: 2.3e-02 relative error at 462², 4.9e-02 at 392², versus 2.7e-06 at 518². The ViT position-embedding interpolation is only partly captured.

So the export **pins 518×518** — which is what the reference preprocessor resizes to anyway (`preprocessor_config.json`: size 518, `keep_aspect_ratio`, `ensure_multiple_of` 14). Pinning costs nothing and removes a silent-wrong-answer mode; the C++ side resizes and optionally letterboxes.

## Parity, because of #1025

The script asserts torch-vs-ORT agreement on noise **and** a smooth structured image (4.1e-06 / 6.9e-06 max abs diff) and refuses to write a graph with any non-finite output. This exists precisely because the UniRig export shipped a 100%-NaN graph that went unnoticed for months (#1025). `PhotoDepth::estimate` carries the same guard at runtime.

## Verification

Model uploaded to the HF models repo (`depth/da2_small.onnx`, 99,111,487 bytes), downloaded through the normal first-use path, and run from the CLI on a 1024×1536 photo — correct depth map, subject near / background far, plausible floor gradient, output at full source resolution.

**12 unit tests** pin the contract: ImageNet normalisation, CHW layout, pinned input size, near=bright, constant-field → mid-grey, non-finite handling, letterbox round-trip, clean failure without a model.

Mutation-tested — inverting near/far, skipping normalisation, mapping a constant field to black, and dropping the per-texel NaN guard each fail a test. One guard was proven **redundant** by mutation (`std::min`/`max` already reject NaN in this argument order) and replaced with a comment explaining why, rather than left as untestable belt-and-braces.

## Not in scope

The issue lists two further consumers — wiring photo depth into the SD ControlNet path as a selectable source, and a height/displacement map for the PBR pipeline. Both are GUI/pipeline work on top of this core and belong in their own PR with their own visual verification.

Closes #1018.

## Summary by CodeRabbit

* **New Features**
  * Added photo-based monocular depth estimation using the Depth-Anything-V2-Small model.
  * Added the `material --photo-depth` command to generate grayscale depth maps for ControlNet workflows.
  * Added optional letterboxing to preserve photo aspect ratios.
  * Depth output uses a near-is-bright convention and matches the source image dimensions by default.
  * The model downloads automatically when needed and is cached locally.

* **Documentation**
  * Added usage, model, licensing, resolution, download, and output-format documentation.


fernandotonon:
All five findings were valid and are fixed in `f3c76e5b`.

**1. Default build broken (Critical) — confirmed and fixed.**

`PhotoDepth.h` sat inside `#ifdef ENABLE_ONNX` while `cmdMaterialPhotoDepth` calls `PhotoDepth::` unconditionally. Your reasoning was exactly right: `PhotoDepth.cpp` always builds and ships non-ONNX stubs *so that* the feature can report itself unavailable, which means the include belongs outside the guard.

Proven both directions by syntax-checking `CLIPipeline.cpp` with `ENABLE_ONNX` stripped from its real compile command (`compile_commands.json`): **4 errors before the fix, 0 after**. I don't have an ONNX-off configure locally, so this was the honest way to verify rather than assert it.

**2. Infinities poisoned the depth range — confirmed and fixed.**

This one corrects a comment I wrote earlier in the PR. I had claimed no `isfinite` check was needed because `std::min`/`max` reject NaN — true for NaN (they return the second argument when a comparison is false), but an isolated `+inf`/`-inf` compares **normally** and takes over the range. Exactly as you describe: `+inf` collapses finite pixels toward black, `-inf` makes the normalisation `inf/inf` = NaN into the 0..255 cast. The inference probe can't catch it because it only rejects output where *every* sampled value is non-finite.

New test pins both signs, mutation-verified (removing the guard fails it).

**3. EXIF orientation — confirmed and fixed, in three places.**

`QImage(path)` reads unrotated pixels, so a portrait phone JPEG was estimated sideways with no metadata on the output to compensate. Now `QImageReader` + `setAutoTransform(true)`.

The finding named the CLI; the same bug was in the **MCP and GUI** paths, so I fixed all three rather than just the one reported.

**4. `--size` could produce an unusable artifact — confirmed and fixed.**

Sharp catch. The graph is pinned to `--size`, but `PhotoDepth::estimate` always builds `[1,3,518,518]`, so a 392-pinned model hosted under the expected filename would fail *every* inference with a dimension mismatch — after passing parity, which makes it worse.

The script now refuses a mismatch against `RUNTIME_SIZE` unless `--allow-size-mismatch` is passed, and says that changing it for real means updating `kInputSize` too. Verified: `--size 392` is refused with no file written.

**5. Stale docstring** — corrected to describe the pinned contract and why it is pinned.

**Re-verified after the fixes:** export still passes torch parity (9.8e-06 noise / 6.9e-06 smooth image), and the CLI still produces a correct 1024×1536 depth map. 13 tests passing.

PR #1028, opened 2026-09-15
feat(#1016): BiRefNet high-quality matting tier (exported, parity-verified, hosted)

### LINKED ISSUE #1016: AI v2 C2: BiRefNet matting upgrade for image-to-3D (Fast/Best tiers)
Part of #818 (Track C2). One of the epic's three named quick wins. Status: **ABSENT.**

`BackgroundRemover` is still U²-Net-only at 320² (`rembg/u2net.onnx`). BiRefNet is MIT (code + weights), gives 1024² mattes with far better edges, and has ready ONNX exports (~930 MB fp32 / ~230 MB quantized; `-lite` variants exist).

## Scope
Drop-in behind the same `BackgroundRemover` interface: add a `Quality::{Fast,Best}` tier (the TripoSR tier precedent), `rembg/birefnet.onnx`, keep U²-Net as the small/fallback tier. New row in `THIRD_PARTY_AI_MODELS.md` in the same PR.

## Acceptance
Side-by-side on 5 test photos: no halo ring in the reconstruction on at least the hair/fur case that fails today.

## Note on scope (measured 2026-09)
A related-sounding matte bug was investigated and turned out **not** to affect the app: the shipped U²-Net path produces clean mattes (13.9% / 16.1% coverage, **0 interior-hole pixels**) on the images that were generating bad faces — see the correction on #998. The threshold-matte failure is scoped to standalone `trellis-cli`. So this issue is a genuine **quality upgrade** (edge fidelity at higher resolution), not a bug fix, and should be scheduled on those terms.

fernandotonon:
Implements **#1016** (epic #818 Track C2) — the epic's first-listed quick win, and the third C-item shipped, which satisfies the epic's "at least three of C1–C6" acceptance criterion.

The model is **exported, parity-verified and hosted** in both the aggregate repo and its own mirror, so this works on first use.

## What ships

`BackgroundRemover::Quality::{Fast, Best}` — U²-Net at 320² (Apache-2.0, unchanged default) and BiRefNet at 1024² (MIT, ~930 MB). Both share ImageNet normalisation and NCHW layout, so the inference path is parameterised by edge length rather than forked.

| surface | how |
|---|---|
| GUI | "High-quality matte (BiRefNet, ~930 MB)" checkbox in AI: Image → 3D |
| CLI | `qtmesh generate3d --matting fast\|best` (best implies `--remove-bg`) |
| MCP | `generate_mesh_from_image { "matting": "best" }` |

Default stays **Fast** everywhere — Best is a ~930 MB download and several CPU-seconds, so it is strictly opt-in and no existing caller changes behaviour.

## BiRefNet is not trivially exportable

Three blockers, each verified rather than worked around:

1. **fp16 checkpoint** vs fp32 tracing input.
2. **`torchvision::deform_conv2d` has no ONNX representation.** I checked whether it was optional first — it is not: the checkpoint carries **150 deformable tensors**, so disabling `ASPPDeformable` would discard trained weights and silently change the model. The exporter decomposes it into GridSample + MatMul, **verified against `torchvision.ops.deform_conv2d` to ~7e-07** on every kernel size BiRefNet uses (1×1, 3×3, 7×7), modulation mask included. The exported graph contains **zero custom ops** — default ONNX domain only, with exactly 20 GridSample nodes matching the 20 deformable convs.
3. **Patching torchvision after import silently does nothing** — `birefnet.py` does `from torchvision.ops import deform_conv2d`, binding the symbol into its own namespace. The script now patches every loaded module and **fails loudly if it patches none**.

Same discipline as the depth export (#1018): a licence gate that refuses anything outside the permissive allow-list *before downloading weights*, pinned input dims with a runtime-size guard, torch-vs-ORT parity assertions (**1.5e-07**), and a refusal to write a graph with non-finite output. Plus one check depth did not need — a **synthetic-disc sanity test**, because a matte that is merely finite can still segment nothing and a constant field would pass parity happily.

## A field bug this PR also fixes

The first GUI build had the tier set but the **Fast model still loaded**, so a 1024² tensor hit U²-Net's fixed 320² graph:

```
Got invalid dimensions for input: index 2 Got: 1024 Expected: 320
```

ORT threw, `removeBackground` returned `ok=false`, and TRELLIS.2 treats that as "no matte available" and carries on — reconstructing the studio backdrop as a flat slab behind the subject. **Ticking the box made output worse than leaving it off.** Fixed at all four call sites so tier and model can never disagree.

It stayed invisible because nothing logged the matte stage, so the stage now reports the model it loaded and shouts on failure:

```
[trellis2] matte: BiRefNet 1024 (birefnet.onnx)
```

Two regression tests pin the contract: a deliberate tier/model mismatch must **fail loudly**, never return an unmatted image that looks like a successful cut-out; and `resolveModelBlocking`'s reported tier must always match the file it returns.

## Measured on a real asset

Full end-to-end pair on a 1254² goblin with a light-grey backdrop, identical settings (TRELLIS.2, balanced, 25k tris, 2048px, seed 42), matte the only variable.

**Matte quality**

| | input | coverage | ambiguous edge |
|---|---|---|---|
| U²-Net | 320² | 13.6% | 2.38% |
| BiRefNet | 1024² | 15.6% | **0.83%** (2.9× crisper) |

**Why the mesh comes out more detailed** — and it is *not* a resolution difference. Both tiers hand TRELLIS.2 an identical **1434×1434** image; the 320/1024 figure is the matte model's internal size only. The cause is how much subject survives:

| | |
|---|---|
| mattes disagree on | 1.96% of pixels |
| BiRefNet keeps, U²-Net **drops** | **32,394 px** (subject U²-Net erased) |
| U²-Net keeps, BiRefNet drops | 719 px |

**45:1 in favour of recovered subject.** U²-Net erases finger tips, ear edges and silhouette detail; TRELLIS.2 can only reconstruct what survives the matte. The Best mesh has defined ear tips, a layered harness with straps and pauldron, and better-articulated fingers.

Aggregate stats look similar (22.8k vs 22.6k verts) because the difference is concentrated at thin extremities — z-range 0.2050 → 0.2210.

## Hosting

`rembg/birefnet.onnx` in the aggregate repo, plus the dedicated mirror [`QtMeshEditor-birefnet-onnx`](https://huggingface.co/fernandotonon/QtMeshEditor-birefnet-onnx) with a full model card (I/O contract, export notes, licence). Both sha256-verified identical.

## Tests

19 `BackgroundRemoverTest` cases. Also fixes a **latent bug** in an existing test: `SegmentsWhenModelPresentElseFallsBack` asserted the output keeps its input size, but with a model present the 0.85 foreground-ratio crop legitimately changes it — it only ever passed because CI has no cached model and always took the fallback branch. Confirmed pre-existing by stashing.

Closes #1016.

## Summary by CodeRabbit

* **New Features**
  * Added optional high-quality background removal for image-to-3D generation.
  * Added UI, command-line, and API options to select standard or high-quality matting.
  * High-quality matting uses the BiRefNet model for improved subject separation.

* **Bug Fixes**
  * Automatically falls back to standard matting when the high-quality model is unavailable.
  * Improved reporting of the matting quality used during generation.

* **Tests**
  * Added coverage for quality selection, fallback behavior, output validity, and model mismatches.


fernandotonon:
Worked through all four findings. Two fixed, one does not apply, one filed separately.

**1. No Fast fallback during GUI preparation (P2) — valid, fixed in `2b2ad0c4`.**

Correct, and the consequence is worse than it first looks: the worker **cannot** download (`ensureModelBlocking` needs a nested event loop, which only the main thread has), so if a user picks Best on a fresh install and the ~930 MB fetch fails while U²-Net is fine, nothing ensures Fast and matting is skipped entirely — reaching the exact backdrop-as-geometry failure this PR exists to fix, by a different route.

Both sites now use `resolveModelBlocking`, which degrades to Fast on the main thread where the download can still happen.

**2. Double Best download for TRELLIS.2 (P2) — valid, fixed in the same commit.**

With TRELLIS.2 and the default "Remove background", the backend-specific ensure was followed by the generic one, so a failing Best fetch could burn **two 40-minute timeouts** before generation even started. A `matteEnsured` flag makes the generic call skip when the backend branch already handled it.

**3. Test could trigger a ~930 MB download (Major) — does not apply.**

`src/test_main.cpp` sets `QTMESH_REMBG_NO_DOWNLOAD` for the whole suite (one of nine such guards), so `resolveModelBlocking` short-circuits before any network access. The analysis looked at the test in isolation and missed the harness.

Verified rather than asserted: I deleted **both** `.onnx` files and ran the test *without* setting the variable explicitly — completes in **0 ms**, downloads nothing, cache still empty afterwards.

The `GTEST_SKIP()` the finding references is also gone for an unrelated reason — CI failed this PR with *"Suite BackgroundRemoverTest produced 2 skipped test(s)"*. This repo treats any skip as a failure (documented in this very file above `SegmentsWhenModelPresentElseFallsBack`, which I'd read past). Both tests now assert real contracts on every runner; confirmed 17 tests / 0 skipped / 0 failures with no models cached, and the same with them present.

**4. No HTTPS enforcement or SHA-256 verification (Major, CWE-494) — valid, but out of scope here → #1029.**

Confirmed: `ModelDownloader` has no scheme or hash validation, and every model base URL is user-overridable. But it is **pre-existing and shared by 21 files**, and this PR does not touch it. Fixing a shared download path under a matting-tier title would change behaviour for 20 other consumers with no test coverage for any of them.

Filed as #1029 with the reproduction, the full consumer list, and a concrete fix (reject non-HTTPS; verify against the HF LFS `oid`, which *is* the sha256 — the same digests #1027/#1028 already used to verify uploads; delete the partial file on mismatch).

30 tests passing locally.

PR #1031, opened 2026-09-15
feat(#1017): LaMa texture inpainting (bake seam fill + paint brush)

### LINKED ISSUE #1017: AI v2 C3: LaMa texture inpainting (bake seam fill + paint brush)
Part of #818 (Track C3). Status: **ABSENT** (no `src/TextureInpaint.*`, no `lama.onnx`; every "lama" grep hit is llama.cpp).

LaMa (Apache-2.0) via `Carve/LaMa-ONNX` / `qualcomm/LaMa-Dilated` — ~200 MB, 512² tiles, reusing `PbrMapSynth`'s feathered tiling.

Three consumers:
1. UV-island seam/bleed fill after `MeshGenBaker` / `MultiViewTextureBaker` (mask = uncovered texels; the baker already knows them from its dilation step).
2. User-facing "Inpaint" brush in texture-paint mode (mask = painted selection).
3. Groundwork for #805's MV-Adapter texture pipeline, which expects an inpaint stage.

New `src/TextureInpaint.{h,cpp}`, `ai_models/pbr/lama.onnx`; surfaces: paint-mode button, `qtmesh material inpaint --mask m.png`, MCP `inpaint_texture`. Licensing row in the same PR.

fernandotonon:
Closes #1017 (epic #818, Track C3).

Fills masked regions of a texture with plausible, seamlessly continued content via **LaMa** (Suvorov et al., WACV 2022 — **Apache-2.0 code AND weights**).

## What's here

New `src/TextureInpaint.{h,cpp}` — Ogre-free + Qt-only (QImage in/out), the `PbrMapSynth`/`BackgroundRemover` shape, so the pure pieces unit-test without a GL context.

The issue's three consumers:

1. **UV-island seam fill.** `MeshGenBaker::Result` now publishes the **pre-dilation** per-texel chart coverage. The dilation pass only smears border colour outward to stop filtering bleed — it does not *continue* the texture — so those are exactly the texels a seam fill should replace. `TextureInpaint::maskFromCoverage` is the bridge.
2. **An "Inpaint" button in texture-paint mode**, filling the current smart selection. It composites the whole layer stack first: an inpainter is context-driven, and the bare active layer would condition it on transparent surroundings.
3. **Groundwork for #805's MV-Adapter.**

**Surfaces:** CLI `qtmesh material --texture  --inpaint --mask <m.png> [-o out] [--mask-dilate N]`, MCP `inpaint_texture`, and Inpaint buttons in the paint panel + detached texture editor. MCP `SERVER_VERSION` → 1.11.0.

## Two silent traps

Both produce a plausible-looking texture rather than an error, so both are pinned by tests — and I verified the tests have teeth by mutation (each mutant fails its matching test, none passed silently):

- **The graph takes the image in `[0,1]` but returns `0..255`.** Feeding an un-scaled image saturates the output — measured mean **246.6 vs a correct 128.6** — with no error raised.
- **The mask is binarised.** The model was trained on a hard 0/1 mask, and passing soft grey through measurably changes the fill.

Unmasked texels are **composited back bit-exact** rather than trusting the model to preserve them. It does (measured leakage MAE `0.0000`), but compositing makes that structural instead of a property we're relying on. Non-finite output falls back to the **source** pixel, so a numerically broken graph degrades to "unchanged" rather than to noise — the #1025 lesson.

## Hosting gotcha

**The `Carve/LaMa-ONNX` repo's plainly-named `lama.onnx` is broken.** It fails ONNX Runtime shape inference on a `DFT` node (LaMa's Fourier convolution) and cannot be loaded at all:

```
Op (DFT) [ShapeInferenceError] one-sided DFT requires real input
```

`lama_fp32.onnx` is the working graph; it's re-hosted under the plain name. `scripts/export-lama-onnx.py` gates on licence *before* downloading any weights, and refuses to host a graph that does not load, emits non-finite values, alters unmasked texels, or leaves the masked region unchanged (a pass-through would satisfy every other check). Verified in both directions — it passes the good graph and rejects the broken one with an actionable message.

## Measured

End-to-end through the CLI on a 768×768 texture (forces the 512² tiling path), masking pixels whose true values are known:

| | |
|---|---|
| texels filled | 24,232 (matches the CLI's own count) |
| reconstruction MAE inside mask | **3.97 / 255 (1.6%)** |
| leakage outside mask | **0.000000** |
| tile seam at x=512 | max column MAE **0.000** |

Models are hosted on both the aggregate repo (`pbr/lama.onnx`, what the runtime downloads) and a dedicated mirror with its own model card. The hosted LFS oid matches the local sha256 of the graph I verified.

## Also

Registers the **BiRefNet** mirror repo, which #1016 hosted but never added to `scripts/sync-hf-model-repos.sh`.

## Test plan

- `TextureInpaint.*` — 14 tests (tensor contracts, both silent traps, non-finite fallback, mask dilation, the coverage bridge polarity)
- `MeshGenBakerTest.*` — 5 tests incl. the new coverage publication
- Full paint/bake sweep: **135 tests pass**
- Real inference verified end-to-end through the CLI (numbers above)

## Summary by CodeRabbit

- **New Features**
  - Added AI-powered texture inpainting for selected regions in the texture editor.
  - Added texture inpainting support to the material CLI and MCP tools.
  - Supports mask dilation, tiled processing for large textures, and preservation of unmasked pixels.
  - The LaMa model downloads automatically on first use when available; clear status messages report progress and errors.
  - Added photo-based depth and depth-letterboxing CLI options.

- **Documentation**
  - Added usage guidance, model details, licensing information, and offline configuration notes.


fernandotonon:
Addressed in 620a2ed. Each finding was verified against the running code first — two were reproduced before fixing, and one optimization was proven equivalent rather than assumed.

**Wire the coverage mask into the bake path** (P1) — correct, and the most important of the set. `MeshGenPredictor` copied `baked.texture` and never touched `baked.coverage`, so the headline seam-fill consumer did not actually run. Now wired behind `Options::inpaintSeams` (CLI `--inpaint-seams`, MCP `inpaint_seams`), **opt-in** because the model is a ~200 MB first-use download, and falling back silently to the dilated bake rather than failing a whole generation over a refinement.

**Avoid waiting after a synchronous download rejection** (P1) — confirmed exactly as described: `startDownload` emits `downloadError` synchronously when busy, so `loop.quit()` fired before `exec()` and was lost. A `settled` flag now skips `exec()` in that case, and only our *own* timeout calls `cancelDownload()` — so a synchronous rejection can no longer cancel the other caller's download.

**Preserve source alpha** (P2 / Major) — reproduced before fixing: a cutout texture with 40,000 transparent pixels came back fully opaque, contradicting this PR's own bit-exactness claim. Fixed at the shared `TextureInpaint` boundary (so CLI *and* MCP both get it), with masked texels forced opaque since invented colour under alpha 0 would be invisible. Re-measured: **942 opacity changes, all inside the dilated mask, zero outside; alpha outside the mask bit-exact.**

**Reset the stroke baseline after inpainting** (Major) — real. `beginStroke()` reuses a non-empty `m_layerStrokeBaseline` as its undo snapshot, so undoing the *next* stroke would also have reverted the inpaint. Cleared, matching every other out-of-stroke layer mutator.

**Bind `runTile` tensors by input name** (Minor) — applied. Agreed it fails deterministically rather than silently, but it fails on the user's machine, and the export verifier can't catch it since it runs the graph with a name-keyed dict.

**Skip inference for tiles with no masked texels** (Major) — applied, and verified to be a *pure* optimization: **max pixel diff 0.000** against the pre-change output, same MAE 3.97 and same 0.000000 leakage, 27% faster (10.1s → 7.4s on the 768² tiling case).

**Validate `mask_dilate` before the blocking download** (Minor) — applied, matching `toolGeneratePbrMaps`/`toolUpscaleTexture`. Verified it now rejects instantly instead of after a ~200 MB fetch.

**Hide the Inpaint button when unavailable** (Minor) — applied to *both* surfaces (the detached editor and the Inspector panel; only the editor was flagged). The header already documented this binding.

---

**Not done, deliberately: moving GUI inference to a worker thread** (Major).

The diagnosis is right — `Qt.callLater` defers, it doesn't offload. But `ensureModelBlocking` runs a nested `QEventLoop` and is main-thread-only by construction, and the sibling `bakeChannel`/`bakePbrSet` actions are synchronous too, so a half-threaded path here would add risk without closing the gap. Instead the blocking is made honest and bounded: a new `inpaintModelPresent()` lets the UI show "Downloading model (~200 MB)…" rather than an unexplained pause, and model acquisition — the long pole at minutes, versus a few CPU-seconds for inference — is now distinguishable from the work itself. Proper threading of the whole paint-action family belongs in its own change rather than being retrofitted onto one button.

PR #1032, opened 2026-09-15
fix(#1023): pass refRoll from GUI and MCP, matching the CLI

### LINKED ISSUE #1023: t2m: GUI 'Generate from text' does not pass refRoll to applyMotionClip (CLI does)
Found while reviewing #1022 (composed motion). Pre-existing on master — not introduced by that PR.

`AnimationMerger::applyMotionClip` takes a final `clipRefRoll` argument: the #954 per-role bind→reference roll that anchors twist to the SOURCE bind pose rather than to whatever the target rig happens to carry.

The **CLI passes it** (`CLIPipeline.cpp`, `clipRefRoll = clip.refRoll` → last argument of the retarget call). The **GUI does not**:

```console
$ git show master:src/AnimationControlController.cpp | grep -c refRoll
0
```

So a clip generated from the Inspector's "Generate from text" box gets a different roll treatment than the identical clip generated from `qtmesh anim --generate` — the GUI silently falls back to the pre-#954 behaviour on libraries that ship `refRoll`.

## Why this is its own issue

Passing it is a one-line change, but it alters the output of **every** GUI-generated clip, so it needs a before/after render check on a couple of rigs (the roll channel is exactly the thing that produced the twisted-arm regressions in #954/#838). That does not belong folded into an unrelated PR.

## Scope
- Thread `clip.refRoll` / `sel.refRoll` through `AnimationControlController::generateMotion` to the `applyMotionClip` call.
- Render a gesture clip (wave/throw — roll shows up on arms) before and after on a Mixamo-style rig and an auto-rigged mesh.
- Check MCP too: `MCPServer.cpp` should be audited the same way in the same pass.

fernandotonon:
Closes #1023.

`AnimationMerger::applyMotionClip` takes a final `clipRefRoll` — the #954 per-role bind→reference roll that anchors twist to the **source bind pose** instead of to whatever reference frame the extraction happened to pick. The CLI has always passed it; **the GUI and MCP never read `refRoll` off the clip at all**, so an identical prompt produced a different roll treatment depending on which surface issued it.

Both call sites also stopped short of `cmuLibraryHandedness`, so this threads both trailing arguments (handedness keeps its existing `true`).

## This is live, not theoretical

All **122 clips** in the shipped `motion-library-v2.json` carry `refRoll`, with up to **1.48 rad (85°)** on body roles. Every GUI- and MCP-generated clip was discarding it.

## Measured before/after

The issue asks for a before/after render check, so here it is — on **two rig styles**, pinning the source clip with `variant_index` so the runs are actually comparable:

**GUI vs CLI, max bone-rotation delta**

| Rig | Clip | Before | After |
|---|---|---|---|
| UniRig-style (Goblin_rigged, 61 joints, unprefixed names) | idle | **70.09°** | 0.06° |
| UniRig-style | death | **28.53°** | 0.05° |
| Mixamo (rumba, 51 bones) | death | **12.16°** | 0.06° |

The residual ~0.05° is float keyframe quantisation in the exported glb, not a behavioural difference — the GUI now reproduces the CLI.

The effect is **far larger on UniRig-style rigs** than Mixamo ones, which is exactly what #954 predicts: the further the target bind sits from the CMU source, the more the roll baseline matters. On the goblin rig the GUI was producing a right leg rotated **70° away** from the CLI's output for the same prompt.

## Methodology note

Pinning matters. `"punch"` resolves to one of three library clips non-deterministically, so an unpinned before/after compares *different source clips* and reports noise. My first measurement did exactly that and showed a misleading 0.05° — the real signal only appeared once the clip was pinned and the generated animation isolated from the four pre-existing Mixamo clips in the same glb.

## Test plan

- Before/after captured through the **MCP surface** (the same `AnimationControlController`/`MCPServer` code paths the Inspector's "Generate from text" uses), on `rumba120/test.fbx` and `~/Downloads/Goblin_rigged.glb`.
- `AnimationMergerTest` failures locally are pre-existing and environmental (`tryInitOgre()` returns false without a GL context) — verified identical on master with the change stashed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved generated motion playback by preserving per-role reference roll data from selected motion clips.
  * Ensured MCP-generated motions apply the same roll handling as other motion-generation workflows.
  * Prevented stitched multi-part motions from incorrectly applying one clip’s reference roll across all segments.
  * Preserved correct reference-roll data for single-clip motions, including repeated single-clip selections.


fernandotonon:
Addressed in 4c35617 — the finding is correct, and I verified the magnitude against the shipped library rather than taking it on faith.

`refRoll` is the per-clip bind→**reference** roll, and the reference frame is chosen per extraction, so takes genuinely disagree:

| pair | disagreement |
|---|---|
| max between any two library clips | **3.94 rad (226°)** |
| run vs punch | 2.13 rad (122°) |
| walk vs wave | 1.01 rad (58°) |
| walk vs sit | 0.56 rad (32°) |

And `"walk then punch"` is a prompt the composer really produces (`generated_walk_punch`, confirmed end-to-end), so this is reachable rather than theoretical.

Two clarifications worth recording:

**It is pre-existing, not introduced here.** The CLI has passed `sel.refRoll` for composed clips since #1010 — so composed CLI clips have been wearing the first take's baseline all along. This PR would have extended that to GUI and MCP, which is exactly why it's the right moment to fix it.

**I did not implement the per-frame version.** `applyMotionClip` adds `refRoll` as one per-role constant over every frame, so correct handling of a stitched clip needs a *per-frame* baseline — a new parameter threaded through all three surfaces. That is a larger change than this PR, and doing it hastily under an unrelated title is how the #954/#838 twisted-arm regressions happened. Instead the composition now emits **nothing** for a multi-take stitch: empty selects the pre-#954 legacy path, which shipped for months, and is strictly better than a known-wrong constant.

This mirrors the gate already applied to **fingers** a few lines above, for exactly the same reason (one take's channel landing on another take's body). `restDir`/`restWorld` are deliberately *not* gated — they describe the source skeleton, which every take shares.

Tests: single-take keeps its baseline, multi-take drops it, and the gate survives the `selectForPrompt` wrapper the surfaces actually call. The fixture now emits a distinct per-clip `refRoll` so "which take's baseline survived" is observable at all — without that the assertions would have been vacuous. Verified by mutation: removing the gate fails two of the three tests.

Filing the per-frame baseline as its own issue.

PR #1035, opened 2026-09-16
fix(#1034): route multi-step prompts to the template path, and say so

### LINKED ISSUE #1034: t2m: 'Use trained model' silently disables multi-step composition (prompt blends into one pose)
Reported from GUI testing.

## Symptom

With **"Use trained model (experimental)"** checked, a sequenced prompt such as:

> `walk then punch, wave twice then sit`

produces **one blended animation** rather than a sequence — the character settles into a sit-like posture and waves/punches *from* it, with the legs shuffling slightly, all simultaneously. Unchecking the box produces the correct four-step sequence.

## Cause

`MotionComposer` is reachable **only on the template path**. In `AnimationControlController::generateMotion` (and the MCP/CLI equivalents) the model branch runs first:

```cpp
if (useModel) {
    const auto mr = MotionGenerator::generate(prompt, mp, ..., duration);
    if (mr.ok) { ...; gotClip = true; }
}
...
if (!gotClip) {          // <- composition lives ONLY here
    ... MotionComposer::selectForPrompt(prompt, lib) ...
}
```

`MotionGenerator::generate` receives the **entire prompt string** and emits a single clip. The CVAE has no notion of sequencing, so "walk then punch, wave twice then sit" is embedded as one text condition and decoded into one averaged motion. The composer never sees the prompt.

So this is not a composition bug — composition is simply bypassed.

## Why it is worth fixing

Nothing tells the user. The checkbox reads "Use trained model (experimental)", which sets expectations about *quality*, not about losing multi-step support. A user writing a sequenced prompt gets a silently wrong result and no diagnostic — the same silent-failure class as #1023.

## Options

1. **Route composition first, model per step.** Parse the prompt with `MotionComposer::parse`; when it yields >1 step, generate each step through the model and stitch the results with the existing `compose` machinery. Best outcome, most work — and the model's per-step quality is still below the template library's.
2. **Fall back to the template path for multi-step prompts** when the model is selected, and say so in the status line ("Multi-step prompt — using the template library; the trained model generates a single motion"). Cheap, honest, matches the current quality ordering.
3. **Disable/annotate the checkbox** when the parsed prompt has more than one step.

Option 2 is probably right for now, possibly with 3 as the UI hint. Whatever ships, the user must be *told* rather than silently handed a blend.

## Notes

- `MotionGenerator` is documented as experimental and below mocap quality (`src/MotionGenerator.h`), so the template path being better here is expected, not a regression.
- Composed generation is also non-deterministic (`MotionLibrary::pickTake` chooses among takes per action) — measured 321 / 223 / 323 frames for the same prompt across three CLI runs. Unrelated to this bug, but it makes before/after comparison of composed clips meaningless unless the clip is pinned; worth its own look.

fernandotonon:
Closes #1034. Found in GUI testing.

## Symptom

With **"Use trained model (experimental)"** checked, `walk then punch, wave twice then sit` produced **one blended animation** — the character settled into a sit-like posture and waved/punched from it, legs shuffling, all at once. Unchecking the box gave the correct four-step sequence.

## Cause

`MotionComposer` is reachable **only on the template path**. The model branch runs first, and `MotionGenerator::generate` receives the *whole prompt* as a single text condition — so the CVAE embeds four action concepts and decodes one averaged motion. The composer never sees the prompt.

```cpp
if (useModel) { ... gotClip = true; }
...
if (!gotClip) {          // composition lives ONLY here
    MotionComposer::selectForPrompt(prompt, lib);
}
```

So this is **not** a composition bug — composition was silently bypassed. I had initially been debugging the stitcher itself; measuring segment poses (120–145° from their solo equivalents) looked like a seam fault, but the code wasn't even executing. The user's observation that unchecking the box fixed it is what identified the real path.

## Fix

All three surfaces consult a new `MotionComposer::promptHasMultipleSteps` before the model branch, route multi-step prompts to the template library, **and say why**:

| surface | how the user is told |
|---|---|
| GUI | status line: "Multi-step prompt — using the template library (the trained model generates a single motion)." |
| CLI | stderr `Note:` |
| MCP | `model_skipped` field in the response |

A caller that asked for the model and got the template must be told, or it reads as the flag being ignored. Nothing in "experimental" suggests multi-step support disappears — that's the defect.

The predicate reuses `parse`'s own `segment()` splitter so the two can never disagree about where a prompt divides, but needs **no `MotionLibrary`**, so the model path can ask before paying for the library download/parse. Deliberately conservative: it counts *segments*, not resolvable actions, so `"walk then flibbertigibbet"` routes to the template path — the one that can actually report the unresolved fragment.

A single action, **including one with a repeat or duration** (`"wave twice"`), is not multi-step and still uses the model.

## Verified end to end

```
CLI  "walk then punch" --model   -> Note printed, composed 2 steps
CLI  "wave" --model              -> (source: model), no note
MCP  model:true, multi-step      -> composed_steps + model_skipped
MCP  model:true, single action   -> source: model, no model_skipped
```

## Scope note

Per-step model generation stitched by the composer would be the better long-term answer, but the model's per-step quality is below the template library's, so routing is the right trade today. Recorded in the issue.

**Unrelated CLI quirk found while testing** (not fixed here): `--model` is only honoured when it appears *after* `--generate` (the `generateMode` gate), so `--model --generate "…"` silently runs the template path. Documented in CLAUDE.md.

## Tests

Pin the predicate against `parse()`'s split, the single-action cases (a false positive there would silently disable the model for every ordinary prompt), and the conservative unresolved-fragment behaviour. Verified by mutation — hardcoding `false` fails two of the three.

## Summary by CodeRabbit

- **New Features**
  - Multi-step text-to-motion prompts now automatically use the template motion library for composition.
  - Single-action prompts—including repeats and durations—continue to use the generative motion model.
  - GUI, CLI, and MCP responses clearly indicate when model generation was skipped and why.
  - CLI model selection is honored when specified after the generation command.

- **Documentation**
  - Updated text-to-motion documentation to explain multi-step prompt handling and model-selection behavior.


fernandotonon:
Good catch — addressed in 0ad85fe. The finding is correct and it defeated the whole point of the PR.

Confirmed the mechanism: the QML handler **assigns** rather than appends —

```qml
function onGenerateMotionStatus(message, isError) {
    genStatus.text = message; genStatus.isError = isError
}
```

— and `generateMotion` emits its success message at the end, so with a cached library the interim notice is overwritten before the event loop can render it. The user saw `(template)` with no explanation of why the checked model option was ignored, which is precisely the silent failure this PR exists to remove.

The reason is now recorded and folded into the **final** message (the one that survives), mirroring the existing `composedNote` suffix pattern a few lines above, plus a `modelSkipped` key on the returned map for programmatic callers:

> Generated 'walk_punch' (template) — 51 bones, 69 frames (2.3s), composed from 2 steps — multi-step prompt, so the template library was used (the trained model generates a single motion)

CLI (stderr, nothing overwrites it) and MCP (`model_skipped` field in the response) were already correct and are unchanged — I re-verified MCP still reports it after this edit rather than assuming the refactor left it alone.

Issue #1037, opened 2026-09-16
ModelDownloader consumers discard downloadError's message — 'refusing http://' and 'SHA-256 mismatch' reach the user as 'offline?'


fernandotonon:
Follow-up from #1029.

All **21** consumers connect `ModelDownloader::downloadError` with the same shape:

```cpp
[&](const QString& name, const QString&) {   // <- error text unnamed, discarded
    if (name == kModelLabel) { ok = false; loop.quit(); }
}
```

and then print their own generic line, e.g. PhotoDepth:

> Error: the Depth-Anything-V2-Small model is unavailable (offline, or QTMESH_DEPTH_NO_DOWNLOAD is set).

#1029 adds two refusals whose *reason* matters to the user — "refusing to download over http://" (a misconfigured base-URL override) and "SHA-256 mismatch, file discarded" (a corrupted or substituted model). The downloader puts the precise text in the signal; every consumer throws it away, so the user is told "offline?" and sent debugging their network.

#1029 mitigates this by emitting both refusals at `qCritical`, which `cliMessageHandler` always prints (it drops `qWarning` unless `--verbose`). That reaches stderr, but the consumer's own error line still says "offline?". This issue is the proper fix.

## Scope

- Thread the second argument through each of the 21 `ensureModelBlocking` handlers into the consumer's error (e.g. `"model unavailable: <downloader reason>"`).
- Audit the GUI path (`AISettingsDialog.qml` calls `startDownload` directly) — confirm the dialog shows `downloadError`'s text rather than a fixed string.
- One shared helper would remove 21 copies of the same lambda; that is the right moment to do it.

Verification bar: set `QTMESH_<X>_MODEL_BASE_URL=http://example.invalid/` with the model absent for two or three consumers and confirm the printed error names the scheme, not "offline".

Issue #1036, opened 2026-09-16
ModelDownloader: resume appends the full body after a stale .part when the server ignores Range (corrupt model that may still LOAD)


fernandotonon:
Found while verifying #1029, reproduced deterministically with a control.

## The bug

`startDownload` resumes from any existing `<dest>.part` by sending `Range: bytes=<size>-`. But **nothing checks that the response honoured it.** The only `Content-Range` handling is progress arithmetic in `onDownloadProgress` (adds `m_resumeOffset` to the totals); `HttpStatusCodeAttribute` (206 vs 200) and the actual `Content-Range` header are never inspected.

So any response that ignores `Range` — `file://` **always** does; any CDN/proxy that strips the header will — returns the **full body with a 200**, and `onReadyRead` appends it after the stale prefix.

## Reproduction (lama.onnx, 208,044,816 bytes)

| stale `.part` | result | |
|---|---|---|
| 29 bytes seeded | **208,044,845 = 29 + 208,044,816**, begins with the seed text | corrupt |
| none (control) | 208,044,816, SHA-256 identical to the real file | clean |

## Why this is worse than a load error

The two corrupted copies behaved **differently**:
- a 30-byte stale prefix → ONNX Runtime: `Protobuf parsing failed` (detected at load)
- a 29-byte prefix → **loaded fine, ran, and produced byte-identical output** to the clean model (ORT parsed the garbage as an unknown protobuf field)

Whether corruption is even *detected* at load depends on what the stray bytes happen to parse as. A successful load proves nothing about integrity — which is the CWE-494 point #1029 makes, and why its SHA-256 check is the actual safety net: with a digest, both copies are deleted instead of cached.

## Fix

In `onDownloadFinished` (or on first `readyRead`), when `m_resumeOffset > 0`:
- read `HttpStatusCodeAttribute`; if it is **200** (not 206) or `Content-Range` is absent/doesn't start at `m_resumeOffset`, the server sent the whole body — **truncate the `.part` to 0 and restart the write** (or discard and re-download), never append.
- for `file://` there is no status; treat resume as unsupported and always start from 0.

Then the stale-`.part` case becomes self-healing instead of self-poisoning. Add a `FakeNetworkReply` test that sets the status attribute to 200 with a non-zero `m_resumeOffset` and asserts the result equals the payload alone.

PR #1038, opened 2026-09-16
fix(#1029): ModelDownloader — refuse non-HTTPS URLs, verify SHA-256 before caching

### LINKED ISSUE #1029: ModelDownloader: no HTTPS enforcement or SHA-256 verification for downloaded models (CWE-494)
Raised by CodeRabbit on #1028 (BiRefNet matting). Filing separately because it is **pre-existing and repo-wide**, not specific to that PR.

## The gap

`ModelDownloader::startDownload` accepts a URL with **no scheme validation** and renames the finished file with **no integrity check**:

```console
$ grep -nE "https|scheme|QCryptographicHash|sha256" src/ModelDownloader.cpp
(no matches)
```

Every model base URL is overridable by the user, via env var or QSettings — e.g. `QTMESH_REMBG_MODEL_BASE_URL` / `ai/rembgModelBaseUrl`, and the equivalent for pbr, unirig, segment, inbetween, motion, t2m, triposr, skintokens, facerig, depth. So a plain-HTTP source or a tampered mirror can substitute any model file, and it is loaded without verification.

CWE-494 (Download of Code Without Integrity Check). Reachability is internal and exploitability difficult — it needs either an attacker-controlled override or a MITM on a plain-HTTP base URL — but the blast radius is every ONNX/GGUF consumer.

## Scope

**21 files** use `ModelDownloader::instance()`. This is not one feature's problem:

```
AIAssistManager, MotionInbetween, PhotoDepth, MeshSegmenter, AIModelCatalog,
UniRigPredictor, MotionGenerator, SkinTokensPredictor, BackgroundRemover, …
```

## Suggested fix

1. **Reject non-HTTPS** in `startDownload` (allow `file://` for local testing, and keep the existing env/QSettings override working for a *trusted* mirror).
2. **Verify SHA-256** before emitting `downloadCompleted` — the digests are already known: the HF API returns each file's LFS `oid`, which *is* its sha256, and several recent PRs (#1027, #1028) verified uploads that way. A per-model manifest of expected digests would make this a one-line check per consumer.
3. Delete the partial file on mismatch rather than leaving it cached, so a corrupted fetch cannot poison later runs.

## Why not fixed in #1028

That PR adds a matting tier; it does not touch `ModelDownloader`. Fixing a shared download path there would mean changing behaviour for 21 consumers under an unrelated title, with no test coverage for the other 20. Better as its own change with its own verification.

Worth noting a related precedent: the UniRig export shipped a graph emitting 100% NaN latents and went unnoticed for months (#1025). Integrity checking at download time would not have caught that (the file was intact, the *export* was wrong), but it is the same class of "we trust an artifact we never verified" problem, and both are worth closing.

fernandotonon:
Closes #1029 — CWE-494 (download of code without integrity check). Pre-existing and repo-wide: **one download path behind all 21 model consumers**, every base URL user-overridable via env/QSettings, no scheme validation, no integrity check.

## Two guards

**1. HTTPS enforcement.** `startDownload` refuses any URL that is not `https://<host>` (or a **host-less** `file://`, for tests and local mirrors — `file://<host>/share/…` is a UNC/SMB remote fetch on Windows and is refused; review caught this). Plain http is a byte-for-byte MITM injection point. Refused **before any filesystem side effect** — no `.part`, no created directory. Every consumer in the tree already uses `https://huggingface.co/…`, so nothing legitimate breaks.

**2. SHA-256 verification.** An optional 4th argument `expectedSha256` (case-insensitive hex) is checked by a local streamed `QCryptographicHash` helper against the finished `.part` **on disk, before the rename**. (Initially reused the updater's `UpdateVerifier::verifySha256Hex`; review caught that `qtmesh_updater` is only linked under `ENABLE_AUTO_UPDATER`, so that would have broken the `-DENABLE_AUTO_UPDATER=OFF` link — fixed, and proven with `nm -u` on the object file.) Never as a running hash in `onReadyRead` — a resumed download appends to bytes this process never saw. On mismatch the `.part` is **deleted** (a poisoned partial must not be resumed from or cached) and `downloadError` fires. Empty digest = legacy behaviour, so the 21 consumers and the two QML call sites in `AISettingsDialog.qml` work unchanged. Wiring digests per model is the follow-up the issue describes (HF's LFS `oid` *is* the sha256).

## Two decisions verified empirically, not assumed

**The scheme refusal is emitted queued, never synchronously.** Every consumer connects its handlers, calls `startDownload`, then `loop.exec()`. A synchronous emit runs their `loop.quit()` *before* `exec()` and is lost — they hang for their full timeout. That's the race the #1017 review caught in `TextureInpaint`, which got a `settled` guard; the other 20 consumers have none. Queued delivery lands inside `exec()` for all of them with zero consumer changes. **Proven on `PhotoDepth` (no guard): an `http://` override fails in 1s instead of hanging 600s.**

**Both refusals log at `qCritical`, not `qWarning`.** `cliMessageHandler` drops warnings unless `--verbose`, and every consumer discards `downloadError`'s text, so a warning would leave the user with "offline?" and no way to learn the cause. Verified: the reason now prints on the CLI with no flags.

## Found along the way — filed, not folded in

**#1036 — resume corrupts the model when the server ignores `Range`.** The `Content-Range` "handling" is only progress arithmetic; nothing checks 206 vs 200. `file://` always ignores `Range`, so the full body gets appended after a stale `.part`. Reproduced with a control:

| stale `.part` | result | |
|---|---|---|
| 29 bytes seeded | **208,044,845 = 29 + 208,044,816**, begins with the seed text | corrupt |
| none (control) | 208,044,816, SHA identical to real | clean |

The alarming part: **the 29-byte-corrupted model loaded, ran, and produced byte-identical output** to the clean one (ORT parsed the garbage as an unknown protobuf field), while a 30-byte prefix earlier failed with `Protobuf parsing failed`. Whether corruption is even *detected* at load is arbitrary. A successful load proves nothing — which is exactly why the digest check in this PR is the real safety net.

**#1037 — all 21 consumers discard the error message**, so the precise reason reaches the user only via stderr today.

## End-to-end (real consumers, real binary)

| scenario | consumer | result |
|---|---|---|
| `file://` mirror, no stale `.part` | inpaint CLI | downloaded, SHA-identical, inpaint OK |
| `http://` override | `TextureInpaint` (has guard) | refused, **2s**, no `.part` |
| `http://` override | `PhotoDepth` (**no guard**) | refused, **1s**, no `.part`, no dir, reason printed without `--verbose` |

## Tests

Plain-http refused with no filesystem side effect · schemeless / `ftp` / `javascript` refused · pure `isAllowedDownloadUrl` predicate agrees with the gate · SHA mismatch never renames and deletes the `.part` · SHA match (uppercase digest) completes · empty digest keeps legacy behaviour.

**Mutation-verified:** disabling the gate fails the refusal test — and lets a *real* network request through in a unit test, which is itself a demonstration of what the gate prevents. Disabling the check fails the mismatch test.

## Summary by CodeRabbit

* **Security**
  * Downloads accept only HTTPS URLs with a host or local file URLs.
  * Remote file paths are refused.
  * Optional SHA-256 verification protects downloaded files before finalization; mismatches are discarded and reported.
  * Rejected download errors no longer expose credentials or query parameters.

* **Bug Fixes**
  * Invalid URLs are rejected before files or directories are created.
  * Download errors are delivered asynchronously through the standard notification flow.

* **Documentation**
  * Updated downloader documentation, including a known resume-download limitation.


fernandotonon:
Correct, and verified rather than assumed — addressed in 406dab89.

`qtmesh_updater` is created and linked only under `if(ENABLE_AUTO_UPDATER)` (`src/CMakeLists.txt:766`), while `ModelDownloader.cpp` is in the unconditional source list, so `-DENABLE_AUTO_UPDATER=OFF -DBUILD_TESTS=OFF` would indeed have failed to link.

Replaced the reuse with a six-line streamed `QCryptographicHash` helper. Coupling every model download to the optional updater target (and its libsodium dependency) was the wrong trade for a hash regardless of the link break; the comment records why the obvious reuse was rejected.

**Proof on the `QtMeshEditor` object file** (cheaper and more direct than a fresh 30-minute configure):

| | `nm -u` undefined refs |
|---|---|
| `UpdateVerifier` before | 1 |
| `UpdateVerifier` after | **0** |
| `QCryptographicHash` after | 4 |

Behaviour unchanged — 28 downloader tests pass, and the mismatch test still fails under a mutant that makes the helper "return" the expected digest.

One honest note: my first attempt at this edit silently changed nothing — the script's final safety assertion tripped on the word "UpdateVerifier" inside my own explanatory comment and aborted before writing. The `nm` count is what caught it. Worth having the object-level check rather than trusting "tests pass".

fernandotonon:
All three addressed in eb557032, each verified rather than assumed.

**`file://` with a non-empty host (Security, Major)** — correct, and the sharpest of the three: `file://server/share/model.onnx` is a UNC path on Windows, i.e. an SMB fetch over the network wearing a local scheme, which walks straight around the https-only rule. `file://` is now accepted only with an empty or `localhost` host, `https://` must carry a host, and malformed URLs are refused. I also made the rejection carry its *reason* — "scheme `file` is not https://" would have been nonsense for this case; it now says the URL would fetch from a remote share. **Mutation-verified:** disabling the host check fails the new UNC test.

**Ignored `QFile::remove` result (CWE-459, Major)** — correct. The scenario is real: a surviving poisoned `.part` plus a later *legacy* no-digest download = an unverified rename. On a failed delete the file is now **truncated to 0 bytes** so any resume starts from byte 0, and the error states which cleanup happened (or that neither worked, naming the path). One honest gap: the failure branch itself isn't unit-tested — forcing `QFile::remove` to fail needs a read-only parent directory, which is a no-op for root on the Linux CI runner and would be flaky.

**Stale `m_expectedSha256` (Minor)** — correct for the tests. Worth noting it was test-only in production terms (every accepted `startDownload` overwrites the member; the tests bypass it by driving `onDownloadFinished` directly) — but I cleared it at both per-download teardown sites (cancel + finish) as defence in depth rather than only in the fixture. Suite verified under `--gtest_shuffle` ×3: 30/30 each run.

fernandotonon:
Valid — addressed in e5c93b9c.

The rejected URL was echoed verbatim into `downloadError`, and that text reaches the GUI status line via `AIModelCatalog`. Since every base URL is user-configurable, `https://user:token@mirror/…?token=…` is a plausible override and its secrets would have been displayed. Userinfo, query and fragment are now stripped via `QUrl::RemoveUserInfo | RemoveQuery | RemoveFragment`; scheme, host and path are kept so the user can still see *what* they misconfigured.

Agreed on the blast radius: `qCritical` already logged only the reason (never the URL) and telemetry reduces the text to a sanitized category, so this was GUI-only exposure — Minor, and cheap to close.

Test asserts the password, token and username are absent from the emitted message while host+path survive; mutation-verified (removing the redaction fails it). Suite 31/31 under `--gtest_shuffle` ×3.

PR #1042, opened 2026-09-16
fix(cli): `qtmesh anim --model` is honoured regardless of position

fernandotonon:
Found while verifying #1034 and documented in CLAUDE.md at the time; this fixes it.

## The trap

`--model` was parsed as `if (arg == "--model" && generateMode)`, so it was **silently dropped whenever it preceded `--generate`**:

```
qtmesh anim rig.fbx --generate "wave" --model   → (source: model)
qtmesh anim rig.fbx --model --generate "wave"   → (source: template)    # trap
```

No diagnostic — the user believed they had exercised the trained model and were looking at template output. Same silent-failure class as #1023 and #1034.

## Fix

`--model` has no other meaning anywhere inside `cmdAnim` (checked), so it is now recorded unconditionally and validated after the loop. Given without `--generate`, it is an explicit error (exit 2, *"--model applies only to --generate"*) rather than being ignored — today the bare flag falls through silently.

## Verified on the built binary

| invocation | before | after |
|---|---|---|
| `--generate "wave" --model` | `(source: model)` | `(source: model)` |
| `--model --generate "wave"` | **`(source: template)`** | `(source: model)` |
| `--model --list` (no `--generate`) | silently ignored | `Error: --model applies only to --generate` · exit 2 |
| `--generate "wave"` alone | `(source: template)` | `(source: template)` |

Two-line behavioural change; the `CLIPipeline*` suites pass unchanged. There is no `cmdAnim`-specific argument-gate test file to extend, and the e2e table above is the stronger evidence for a parsing fix.


fernandotonon:
Valid — addressed in 834563b4. The Alembic `<clip>.abc --info` branch returned before the new validation ran, so `--model` was still silently accepted on that one path — the exact gap this PR claims to close. Moved the check to directly after the argument loop, so it fires on every path.

Verified on the built binary: `anim /x/clip.abc --info --model` → exit 2 with the diagnostic; `anim /x/clip.abc --info` (control) still reaches the Alembic branch; the four original invocations are unchanged.

Issue #521, opened 2026-05-14
Anim: Slice D — Pose library (named poses, blend, apply, export)


fernandotonon:
Parent epic: #517

## Goal

A **named pose** is a frozen snapshot of bone TRS (and optionally morph weights). Authors create pose libraries (T-pose, A-pose, neutral, "smile_l", "smile_r"), apply them with one click, and blend between them. Standard Blender / Maya / Houdini tooling.

## Scope

- **`PoseLibrary`** singleton per entity (or per skeleton, shared across entities sharing a skeleton).
- Persist poses in the project file + a sidecar `.poselib` JSON for easy sharing/version control.
- Operations:
  - **Save current pose**: capture bone TRS + active morph weights, prompt for name.
  - **Apply pose**: snap to the pose (with optional duration → blend in over T seconds).
  - **Blend two poses**: dual-quat-correct slerp between A and B at user weight; result becomes the live pose.
  - **Mirror pose**: flip a pose across the symmetry axis (heuristic bone-name mapping — `_l`/`_r`, `Left*`/`Right*`).
  - **Delete / rename**: standard list management.
- **GUI**: a "Pose Library" panel in Animation Mode with thumbnails (auto-rendered when a pose is created — reuse `MaterialPreviewRenderer` infra).
- **Apply-with-mask**: only apply to a selected bone subset (e.g. apply a facial expression without disturbing the body pose).
- **Round-trip**: poses save into the project; export option to bake poses into named one-frame skeleton animations (so DCC tools that don't have a pose-library concept still get the data).
- **CLI**: `qtmesh pose model.fbx --library list`, `qtmesh pose model.fbx --library apply smile_l -o out.fbx`.
- **MCP**: `save_pose`, `apply_pose`, `list_poses`, `delete_pose`, `mirror_pose`.

## Acceptance Criteria

- [ ] User can save the current pose under a name and reapply it later in the session and across reloads.
- [ ] Apply with blend duration smoothly interpolates from current pose to target.
- [ ] Mirror correctly flips a left-side facial expression to right side on a fixture mesh with `_l`/`_r` naming.
- [ ] Apply-with-mask leaves unselected bones unaffected.
- [ ] Pose thumbnails render in the panel.
- [ ] Round-trip: save project, close, reopen — pose library intact.
- [ ] All operations are undoable.
- [ ] Sentry breadcrumb `scene.anim.pose.*` recorded.

## Effort

~7 days.

[...]

PR #1055, opened 2026-09-18
fix(#521): make applied poses actually render + Pose Library GUI/MCP fixes

fernandotonon:
## Summary

Slice D2 shipped the Pose Library UI, but **applying a pose did nothing visible** — it blinked for a frame and reverted, and the number of clicks before it "took" was random (3, 5, sometimes 20). Saving a pose also dragged the model to the world origin and cleared the selection.

All fixed and **confirmed working in the GUI by the reporter.**

## Root cause: the bones were always correct

A temporary instrumented build logging every apply showed the write path succeeding on **every** click — 62 bones written and held, no release, no reset. The pose never reached the GPU:

- The skin samples `Entity::mBoneMatrices`, refreshed only by `cacheBoneMatrices()`, gated on *"frame number changed OR `Skeleton::getManualBonesDirty()`"*.
- With `setSkipAnimationStateUpdate(true)` suppressing the per-frame re-apply, that dirty flag is the **only** thing that re-uploads the pose.
- `flushSkeletonPose()` called `Skeleton::_updateTransforms()`, which **clears** the flag — so new matrices were never uploaded and the mesh kept drawing the stale pose until some unrelated event happened to dirty it. Hence the randomness.

`flushSkeletonPose` now sets `_notifyManualBonesDirty()` **after** the `_update()` calls and lets `_updateAnimation()` consume it. The ordering is the fix.

## Holding a pose

`holdPosedBones` / `releasePosedBones` use both halves of the recipe the bone-drag gizmo already relies on (`TransformOperator::mousePressEvent`):

1. `Bone::setManuallyControlled(true)` — excludes the bone from `Skeleton::reset()`, which would snap it to BIND (the "everything goes T-pose" symptom).
2. A zeroed per-state **blend-mask entry** — manual control alone does *not* stop animation tracks applying each frame, the "pose flashes then reverts" symptom.

Doing only (1) breaks reset semantics: `reset(false)` *skips* manual bones, so a re-enabled clip lands on an unreset bone and the error compounds on every toggle.

Holds record exactly what they overwrote (owned handles, prior `skipAnimationStateUpdate`, prior per-clip mask weights, which masks we created) so release restores precisely that — other systems (the gizmo, mocap) own blend masks that must not be trampled.

Released on **Play** (every pose-held entity in the scene, not just the selection) and after a **thumbnail** render.

## Other reset paths guarded

- The undo/redo refresh in `MainWindow::initToolBar` ran `reset(true)`, which discards manual control.
- That handler checked only `AnimationControlController`'s entity, while the Pose Library poses `SelectionSet`'s — they differ until a clip is picked, which is why the *first* apply after a fresh selection behaved differently from later ones.
- `tickBlend` writes bones itself and got the same hold+flush; without it every eased frame was written but never rendered. Covers the GUI "Blend in" duration and MCP `apply_pose` with `duration`.
- `clearAll()` wiped every container **except** `m_holds`, leaking a hold past the reset that should have dropped it.

## Capture renderers mutated live editor state

`ModelTurntableRenderer` was written for a CLI process that exits right after rendering, so it freely mutates global state and never restores it. The Pose Library renders thumbnails against the **live scene**, which surfaced three instances:

| Symptom | Cause |
|---|---|
| Model jumped to the world origin on save | `recenterEntitiesAtOrigin` never restored node positions → `RecenterGuard` |
| Selection cleared on save | `prepareSceneForCapture` calls `SelectionSet::clear()` → `SelectionGuard` |
| Thumbnail cancelled an applied pose | unconditional `releasePosedBones` → only releases a hold it created |

`ModelIsometricRenderer` had the identical selection bug and **is** reachable in-editor via `IsometricSpritesController`, so "Export Isometric Sprites…" is fixed too.

If a fourth instance appears, the durable fix is one capture-session guard rather than a guard per piece of state.

## QML

- `PosePicker` is an **inline component**, and Qt inline components don't share the enclosing component's scope — so `poseSection.poseNames` never resolved and both the enabled-check and the ListView model were dead bindings. This was the real reason the "Pose to apply" dropdown was empty. Now passed in via an explicit `names` property.
- The popup delegate used a bare `modelData` under `pragma ComponentBehavior: Bound`, so rows rendered blank and selected `undefined`.
- The mask bone list used a stock Qt `CheckBox` instead of the Inspector theme.

## MCP

All pose tools (`apply_pose`, `save_pose`, `mirror_pose`, `blend_poses`, `apply_pose_masked`, `delete_pose`, `list_poses`, `save_pose_library`, `load_pose_library`) delegate to `PoseLibrary`, so they inherit every fix above. Auditing that surfaced the `tickBlend` gap fixed here. Additionally:

- `duration > 0` is now **rejected in headless `--mcp`**: the blend is advanced only by `MainWindow::frameRenderingQueued`, so with no GUI it reported success and left the rig mid-transition forever.
- `duration` / `weight` are type-validated; previously a non-numeric value silently became `0.0` / `0.5`.

## Tests

Nine new tests, all passing in CI:

- apply survives an `_updateAnimation` tick
- save captures the clip-driven pose rather than bind
- a blended apply holds its bones mid-transition and converges on the target
- a thumbnail leaves an unheld posed rig untouched
- save/overwrite/delete leave no hold behind
- release preserves a foreign blend-mask weight
- `forgetEntity` drops an in-flight blend + hold
- the turntable restores node position **and** selection after rendering

## Review

13 findings across two reviewers, each verified against the code before fixing: **Codex 5** (thumbnail discarding the restored pose, commands leaking holds, dangling entity on destroy, Play releasing only the selection, release clobbering foreign masks) and **CodeRabbit 6** (QML inline-component scope, test resource-name collision, headless timed apply, numeric validation, masked apply not cancelling a blend, stale comment).

The test lane then caught two more of my own: the `m_holds` leak in `clearAll()`, and a thumbnail test that passed for the wrong reason because it never selected the entity.

## Notes for reviewers

Two things cost real debugging time and are worth knowing about this codebase:

- `QT_NO_WARNING_OUTPUT` / `QT_NO_DEBUG_OUTPUT` are defined in this build config, so `qWarning`/`qDebug` **compile to nothing**. Diagnostics added that way vanish silently.
- The app detaches from the launching shell's stdio, so shell-level stderr redirection captures nothing from a bundle launch. File-based logging from inside the process is what works.


fernandotonon:
@coderabbitai review

fernandotonon:
Thanks — all five findings were real, and I verified each against the code before fixing. Addressed in 221ce5bb.

**P1 — thumbnail discarded the restored pose.** `releasePosedBones` ran *after* `applySnapshot(entity, before)` and did `reset(true)`, throwing away the TRS just restored. Release now happens *before* the restoring write, and `releasePosedBones` no longer resets at all — callers write the state they want kept immediately beforehand.

**P1 — commands leaked holds.** Routed all eight transient-restore sites (`SavePoseCommand`, `DeletePoseCommand`, `capturePriorPose`, and the apply-undo paths) through a new `restoreAndUnhold()`.

**P1 — dangling entity.** Confirmed `forgetEntity` had zero production callers; hooked it into `Manager::destroyAllAttachedMovableObjects` before the entity is destroyed. It now drops the `m_holds` entry too — a stale entry would make `hasHeldBones()` true for a dead pointer.

**P1 — Play released only the selection.** `setPlaying` now releases every pose-held entity in the scene.

**P2 — release clobbered foreign blend masks.** Good catch on the `TransformOperator` precedent. Holds now record the prior weight per pose-owned handle and restore only those; masks the hold itself created are destroyed rather than left at full weight.

This meant making `holdPosedBones`/`releasePosedBones`/`hasHeldBones` instance methods backed by an `m_holds` map recording, per entity: owned bone handles, prior `skipAnimationStateUpdate`, prior per-clip mask weights, and which masks we created.

Four regression tests added: a thumbnail leaves an unheld posed rig untouched; save/overwrite/delete leave no hold behind; release preserves a foreign mask weight; `forgetEntity` drops an in-flight blend + hold so `tickBlend` can't touch it.

fernandotonon:
Thanks — all six were real. Fixed in 7062acf3.

**The inline-component scope catch is the important one.** You're right that `PosePicker` never had access to `poseSection`, so both the enabled-check and the ListView model were dead bindings. That — not the `modelData` fix in the first commit — is the actual root cause of the empty "Pose to apply" dropdown the user reported; the earlier fix was necessary but not sufficient. Now passed in via an explicit `names` property, bound at all three use sites.

**The test resource-name collision would have failed CI.** `PoseLib_Forget` was already taken at line 204, and `createAnimatedTestEntity` registers Ogre resources by name that `TearDown` doesn't remove. Renamed to `PoseLib_ForgetBlendHold`.

**Headless timed apply**: good catch that `tickBlend` has exactly one non-test caller. `toolApplyPose` now rejects `duration > 0` when no `MainWindow` is attached rather than reporting success and leaving the rig mid-transition.

**Numeric validation** on `duration`/`weight`, **`applyPoseMasked` now cancels an in-flight blend** (same contract `applyPose` already had), and the **stale scope comment** in `MeshImporterExporter.cpp` is gone — scene-level persistence shipped in 5f7472c4.

Issue #1019, opened 2026-09-14
AI v2 C1: Audio2Face-3D lipsync (audio → ARKit blendshape weights)


fernandotonon:
Part of #818 (Track C1). Status: **ABSENT** (no `src/AudioToFace.*`, no `a2f/` models, zero viseme/lipsync hits).

NVIDIA open-sourced Audio2Face-3D in Sept 2025 (MIT SDK + open weights). Audio → ARKit blendshape weights over time.

This plugs into infrastructure that already exists end-to-end: #889 FaceRig generates the 52 ARKit targets on any humanoid face, and #869 mocap already writes blendshape weight keyframes via `MorphAnimationManager::writeWeightKeyOn`. So the new work is the audio→weights predictor plus surfaces; the rigging and keyframe-writing halves are done.

Surfaces: CLI (`qtmesh lipsync take.wav --mesh head.glb`), MCP, and a GUI entry alongside Performance Capture. Licensing row in the same PR; verify the exact weight files, not just the SDK.

fernandotonon:
-

fernandotonon:
-

fernandotonon:
-

fernandotonon:
## Licence check done — A2F is usable, Audio2Emotion is not

The epic flagged "verify the specific regression-model card at implementation time". Done, and it splits in two:

| component | licence | usable |
|---|---|---|
| Audio2Face-3D models | NVIDIA Open Model License | yes |
| **Audio2Emotion models** | **"Custom (use allowed with Audio2Face only)"** | **no** |

The [NVIDIA Open Model License](https://www.nvidia.com/en-us/agreements/enterprise-software/nvidia-open-model-license/) clears our bar — "Models are commercially usable", with a grant to "reproduce, use, create derivative works ... distribute". So the A2F weights can go on the download-on-first-use path like every other model.

It is not obligation-free, and these need honouring at implementation time:

- **Redistribution requires a NOTICE file.** "If you distribute the Model, You must give any other recipients of the Model a copy of this Agreement and include the following attribution notice within a 'Notice' text file with such copies: 'Licensed by NVIDIA Corporation under the NVIDIA Open Model License'".
- The licence is **revocable**, and auto-terminates if you "bypass, disable, reduce the efficacy of, or circumvent any technical limitation, safety guardrail", or on patent/copyright litigation against NVIDIA.
- Use must stay "consistent with NVIDIA's Trustworthy AI terms".

[...]

Issue #889, opened 2026-07-14
Epic: AI — Auto-generate ARKit blendshapes on any humanoid mesh (deformation transfer)


fernandotonon:
## Overview

Automatically attach the **52 ARKit-style blendshape morph targets** to an arbitrary humanoid face/head mesh — the same "one-click auto-rig" idea as [#407 AutoRig](https://github.com/fernandotonon/QtMeshEditor/issues/407) / #408 UniRig / #819 SkinTokens, but for the **face** instead of the skeleton. Input: a neutral head with no blendshapes. Output: the same mesh carrying `jawOpen`, `mouthSmileLeft`, `eyeBlinkRight`, `browInnerUp`, … so that **face performance capture (#869) works on any suitable mesh** without the user having to source a Ready Player Me / CC4 head or hand-sculpt 52 shapes.

This removes the single biggest friction point for the face-mocap feature ("I need a special mesh with ARKit targets").

## Technique — deformation transfer (NOT an ML model)

The standard, well-published approach is **Deformation Transfer** (Sumner & Popović, SIGGRAPH 2004):

1. A **template** head that already has the 52 ARKit shapes.
2. **Non-rigid ICP (NRICP)** fits the template to the user's neutral mesh (correspondence + warp).
3. For each of the 52 expressions, transfer the template's per-triangle deformation onto the user's topology → a new morph target on the user's mesh.

Crucially this is a **deterministic geometry algorithm (sparse linear solve), not a neural net** — so unlike UniRig/SkinTokens it needs **no ONNX, no model download, no worker-thread inference**. It slots in beside the native `GeodesicVoxelBind` / `QuadRetopo` implementations.

## Licensing (the deciding factor, as always)

- **Algorithm:** Deformation Transfer paper is unencumbered; the reference implementation [vasiliskatr/deformation_transfer_ARkit_blendshapes](https://github.com/vasiliskatr/deformation_transfer_ARkit_blendshapes) is **MIT**.
[...]

PR #1057, opened 2026-09-19
fix(#889): ARKit blendshapes — wrong-side correspondence, unconverged solve, and mismatched prealign

fernandotonon:
The user reported that the generated ARKit blendshapes give poor results *even on template-like face meshes*. I diagnosed before building, because **Audio2Face (#1019) only outputs weights over these shapes** — it would have inherited the problem and been blamed for it. Full diagnosis posted on #1019.

## The template is not at fault

Parsed `arkit_template.bin` (ICT-FaceKit, 26719 verts, 51 shapes) directly:

- **Every shape deforms.** Magnitudes are anatomically sensible — `jawOpen` largest at 8.8% of the head diagonal, `eyeLook*` smallest at ~0.5%.
- **All 20 Left/Right pairs are genuinely mirrored**, not duplicated, so one-sided expressions are representable.

## The loss was in the resample

Each user vertex took the delta of **one nearest template vertex** — topology-blind at exactly the places that matter:

| measurement | value |
|---|---|
| median template vertex spacing | 0.59% of head diagonal |
| adjacent verts at the lip seam under `jawOpen` | one moves **1.6**, the other **exactly 0**, `0.054` apart |
| query offset by half a median edge → motion changes drastically | **3.2%** of verts near a seam |

Upper and lower lip touch in the neutral pose but only one follows the jaw, so a nearest-vertex pick there is a coin flip. A scattered slice of lip and eyelid vertices inherited the opposite side's motion, or none — sparse enough to read as "mushy expressions" rather than a bug.

## The fix

`SurfaceSampler` projects each query onto the nearest **triangle** of the correspondence mesh (Ericson region test, clamped to edges/corners) and blends the three corner deltas barycentrically. The corners belong to one triangle, so a query on the upper lip can only mix upper-lip motion — it cannot cross the seam, because no triangle joins the lips.

## Measured result

On a retessellated head (same surface, vertices slid up to 45% along their edges — the realistic "user mesh is a different tessellation" case):

- **3.7% of vertices got a different delta**, matching the 3.2% predicted from the template analysis.
- The changed shapes are exactly the mouth group: `cheekPuff`, `mouthPucker`, `mouthSmile`, `jawOpen`, `mouthFunnel`.
- `jawOpen`'s largest single correction is **1.13 on a shape of scale 0.75** — a vertex that had been taking the wrong side of the lip entirely.

Both versions report an **identical fit residual** (0.0207% mean, 0.8098% max). That is the point: the existing gate measures the *neutral* match and cannot distinguish a good rig from one with inverted lip motion.

## Second defect, found by testing the fix: shapes were 4-9x TOO WEAK

The correspondence fix alone did not make the rig usable. Testing on the **ICT template as its own control** (fit residual 0.014% — as close to a perfect fit as this pipeline gets, so anything wrong is downstream of fitting) every shape arrived far too weak:

| shape | ratio vs template |
|---|---|
| `mouthSmileLeft` | **0.11** |
| `jawOpen` | **0.14** |
| `cheekPuff` | 0.16 |
| `eyeBlinkLeft` | 0.26 |

That is the reported "barely moves". Two independent causes, both measured:

### 1. The CG solve never converged, and failed silently
Plain CG on the normal equations squares the condition number, and this system mixes ~156k per-triangle gradient rows with a handful of anchor rows. On a 27k-vertex head the **800-iteration cap was exhausted every single time**. The result was not merely scaled down but **distorted** — per-vertex amplitude scattered **0.22-1.41** of target.

The residual looked respectable (~0.018), which is why this hid: most rows satisfy early while the low-frequency modes — the ones carrying overall amplitude — are still converging.

| iters | per-vertex ratio (median / p10 / p90) |
|---|---|
| 800 | 0.455 / 0.218 / 1.413 ← shipped |
| 4000 | 0.781 / 0.364 / 0.898 |
| 8000 | 0.998 / 0.935 / 1.003 ← converged |

Fixed with a **Jacobi preconditioner** (`diag(AᵀA)`, O(nnz) to build, one divide per iteration). The solve now **exits on tolerance at ~4.6k iterations** instead of burning the cap, so the rig got *faster* as well as correct.

### 2. The prealign compared different subsets of the head
NRICP pre-aligns by centroid + bbox diagonal, but the template side is the **main surface only** (the component split removes eyeballs/teeth/lashes) while the user side is the whole head-masked mesh *including its own equivalents*. On the template fitted to **itself** the two centroids differed by **0.875 in Z**, so the anneal warped the surface by ~0.55 mean to close a gap that should not exist — detuning the transfer's target rest frames.

`NricpOptions` gains `prealignTmplV`/`prealignUserV` so the caller aligns like with like. Drift fell **0.747 → 0.197** mean.

### Result
All 51 shapes on the control mesh: **median ratio 1.00** (min 0.98, max 1.02), **51/51** within tolerance. `mouthSmileLeft` went 0.11 → 1.00.

### Test
`DeformationTransfer.CurvedIdentityFitPreservesAmplitude`, **mutation-verified** — it fails with either half of the fix reverted. Sized at ~10.2k verts deliberately: with the fix reverted the ratio is 1.000 at 441/1681/3721/6561 verts and only breaks at 10201, so the existing flat 10×10 grid test could never have caught this. It asserts the **per-vertex** distribution (p10/p90), not just the peak, because an unconverged solve distorts rather than uniformly scaling.

## Follow-ups (not in this PR)
- **An expressiveness gate** — per-shape displacement against the template's own profile. Today nothing catches "attached but wrong", only "attached but invisible" (the existing amplitude retry).
- **Region masks** so a query can never borrow across a crease even where the surface is folded tightly. The barycentric projection removes the common case; masks would close it structurally.
- **Then #1019 Audio2Face**, which becomes a clean win — `MorphAnimationManager::writeWeightKeyOn` already writes the keyframes.

Tests: existing FaceRig/NRICP/DeformationTransfer suites pass (14). Verified end-to-end through the real `qtmesh facerig` CLI on a 26719-vert head, 51 shapes attached both before and after.

## Summary by CodeRabbit

* **New Features**
  * Added SDXL Base 1.0 as an available image-generation model.
  * Added adjustable FLUX.2 generation steps, from 4–20.
  * Added “Save Image As…” for generated or loaded source images.
  * Clarified image editing versus creating a new image.
  * Updated image-model terminology across settings.

* **Bug Fixes**
  * Improved generated image sizing and subject framing for SDXL and FLUX.2 models.
  * Improved 3D shape transfer accuracy and deformation quality.

* **Changes**
  * Removed SDXL Turbo from the model catalog because its license does not permit commercial use.


fernandotonon:
-

PR #1064, opened 2026-09-19
ci(winget): sync the winget-pkgs fork before submitting

fernandotonon:
The 3.40.1 release failed in `winget-publish` ([run 35430677202](https://github.com/fernandotonon/QtMeshEditor/actions/runs/35430677202)). Every other job in that release succeeded.

## What failed

The manifest generated and validated fine (`Manifest validation succeeded: True`). The submit is what failed:

```
Submitting pull request for manifest...
The forked repository could not be synced with the upstream commits. Sync your fork manually and try again.
```

`fernandotonon/winget-pkgs` was **0 ahead / 1627 behind** `microsoft/winget-pkgs`:

```console
$ gh api repos/fernandotonon/winget-pkgs/compare/microsoft:master...fernandotonon:master \
    --jq '{ahead:.ahead_by, behind:.behind_by}'
{"ahead":0,"behind":1627}
```

`wingetcreate` branches the new manifest off the fork's default branch and refuses when that branch is stale. The job's three retries could never have worked — the state is persistent, not transient. The old error text did list a diverged fork as possibility 2 of 3, but nothing acted on it.

## The fix

Sync the fork over the REST `merge-upstream` endpoint before submitting — a fast-forward, so it cannot discard fork-local commits.

**That sync does not always succeed, and the job now handles that honestly.** Upstream regularly edits `.github/workflows/**`, which a classic `public_repo` PAT may not write, so `merge-upstream` returns 422:

```
refusing to allow an OAuth App to create or update workflow
`.github/workflows/domain-validation-assist.lock.yml` without `workflow` scope
```

I confirmed this is persistent, not incidental: two attempts an hour apart failed identically while the fork drifted from 1627 to 1633 behind. Since `wingetcreate` cannot branch from a stale fork, the job re-reads the comparison after a failed sync and exits immediately if the fork is still behind, naming the two one-time fixes — press **Sync fork** on the fork's page, or add `workflow` scope to `WINGET_TOKEN`. Failing in seconds with the remedy beats three retries that cannot succeed.

A fork that is *ahead* of upstream is deliberately not rewritten; that needs a human decision, and the final error says to reset it.

The 422 status is read defensively: PowerShell Core throws `HttpResponseException`, whose `.Response` can be null, so reading `.StatusCode.value__` off it unguarded would throw inside the `catch` and mask the real error.

## Before this merges

**The fork still needs one manual sync.** This PR stops future releases from silently burning retries, but it cannot perform the workflow-scope sync itself. Press **Sync fork** at https://github.com/fernandotonon/winget-pkgs once, and 3.40.2's winget job should go green.

## Merge order

Land this **before** tagging 3.40.2 (#1063). `winget-publish` runs on `release: published` and checks out the default branch, so it picks up whatever is on `master` at publish time.

## Testing

YAML parses; brace/paren/bracket balance verified across the script excluding comments and string literals. The sync path itself cannot be exercised without publishing a release — 3.40.2 is the first real test.


fernandotonon:
Superseded: #1062 merged while this was being raised, so this branch conflicted. Re-raised with only the follow-up commit rebased onto master.

PR #1069, opened 2026-09-21
feat(#1058): contextual feedback prompt to explain one-session churn

### LINKED ISSUE #1058: Add lightweight contextual feedback for key user workflows
## Goal

Understand why users successfully open/import models but often do not export or return to QtMeshEditor the following week, without creating an intrusive feedback system.

We already measure acquisition and weekly retention, but telemetry alone cannot explain why users churn. This feature should add lightweight contextual feedback at key workflow moments so qualitative feedback can be correlated with activation and retention.

Keep this deliberately small and suitable for an MVP. **Do not turn it into a large survey, account system, or analytics platform.**

## Feedback triggers

Show contextual feedback after meaningful workflow moments:

- First successful import
- Import failure
- First successful export
- Export failure
- Extended/meaningful session without an export

The exact threshold for an "extended/meaningful session" can be simple/configurable in the first implementation.

## UX

- Maximum **1 feedback prompt per session**.
- Keep the initial interaction extremely lightweight:
  - Positive
  - Negative
  - Dismiss
- If negative, optionally ask what went wrong.
- Suggested categories:
  - Materials
  - Textures
  - Animation
  - Skeleton / rigging
  - Geometry
  - Scale / orientation
  - Performance
  - Something else
- Allow an optional short text comment.
- Do not repeatedly ask users who dismiss the prompt.
- Implement cooldown using persistent settings / `QSettings`.
- Provide an option such as **"Ask for occasional feedback"** that can be disabled.
- Feedback UI must not block or interrupt normal editing.

## Privacy

Never upload:

- Model contents
- Filenames
- Filesystem paths
- Textures/assets
- Other user content

Allowed contextual metadata can include:

- Anonymous installation ID
- QtMeshEditor version
- OS/platform
- File format
- Importer/exporter used
- Error category
- GUI/CLI workflow
- Session duration
- Workflow stage

Prefer coarse/non-sensitive values where possible.

## Telemetry

Record Sentry events such as:

- `feedback.prompt_shown`
- `feedback.dismissed`
- `feedback.positive`
- `feedback.negative`
- `feedback.submitted`

Include useful non-sensitive context so feedback can later be correlated with activation and retention, for example:

- anonymous installation ID
- application version
- platform
- workflow stage
- file format
- importer/exporter
- error category
- GUI/CLI
- session duration
- feedback category, when provided

Do not attach model/user content to Sentry events.

## Backend

Keep the first implementation minimal.

If a backend endpoint is required, something like:

```
POST /api/feedback
```

is enough.

Requirements:

- Anonymous
- Rate-limited
- Minimal payload
- Offline-safe
- Fail silently on network/backend errors
- Feedback infrastructure must **never interrupt editing or make a workflow fail**

If Sentry alone is sufficient for the MVP, avoid adding a backend until one is actually needed.

## Product questions this should help answer

After implementation, we should be able to start answering questions such as:

- Did the user accomplish what they wanted?
- Where did unsuccessful workflows fail?
- Why do users import a model but not continue/export?
- Are particular formats or workflows associated with negative feedback?
- Can feedback help explain low W1 retention?

## MVP acceptance criteria

- [ ] Feedback can be triggered after the defined meaningful import/export/session events.
- [ ] At most one prompt is shown per session.
- [ ] Initial prompt supports positive / negative / dismiss with minimal friction.
- [ ] Negative feedback can optionally include one category and a short comment.
- [ ] Dismissal/cooldown state is persisted using `QSettings` or equivalent.
- [ ] Users can disable occasional feedback prompts.
- [ ] No filenames, paths, assets, model contents, textures, or other user content are uploaded.
- [ ] Sentry records the feedback lifecycle events with non-sensitive contextual metadata.
- [ ] Feedback submission is anonymous and rate-limited if a backend endpoint is introduced.
- [ ] Feedback collection is offline-safe and fails silently.
- [ ] Editing/import/export workflows never depend on feedback infrastructure being available.
- [ ] The resulting data can be correlated at an anonymous installation level with activation/retention telemetry.

## Non-goals

- Large surveys
- Mandatory feedback
- User accounts/authentication
- Full analytics platform
- Session replay
- Uploading project/model contents
- Complex backend infrastructure
- Prompting on every workflow action

fernandotonon:
Closes #1058. Editor half — depends on **qtmesh-cloud#107** for the anonymous endpoint.

## The question this exists to answer

Users open the editor once and don't come back. Retention data can't distinguish **"they got what they came for"** from **"they got stuck"** — identical in the numbers, opposite responses required. So the prompt is built to separate those two, not just to detect failure.

That shapes the main UX decision: the prompt offers **two affirmative answers**, not a thumbs up/down.

> **Got what I needed** · **Something didn't work** · **Not now**

A satisfied one-time user and a blocked one are both "didn't return" in the data; here they're distinguishable.

## Triggers

| Trigger | When |
|---|---|
| `first_export` | First ever successful export — the clearest "did this work for you?" moment |
| `import_failure` / `export_failure` | Something broke |
| `session_no_export` | Imported, worked ≥12 min, never exported — the churn-shaped case |

**A successful first import deliberately does *not* prompt**, despite being listed in the issue. It's the *start* of the workflow: asking there interrupts someone who hasn't yet had the chance to succeed or fail, and burns the one-per-session budget before the informative moment arrives.

## Policy

- At most **one prompt per session**
- **14 days** between prompts; stops entirely after **3 consecutive dismissals**; answering resets that budget
- `maybePrompt` checks `isSignalConnected(promptRequested)` before consuming the session budget, so **headless CLI/MCP runs never burn it** (the `GamificationManager::maybeRequestConsent` guard)
- Non-modal `QMessageBox` — never blocks editing. Negative hands off to the existing `FeedbackDialog`, prefilled from the trigger.

The session trigger runs on a **60s timer, not in `closeEvent`** — that path calls `QApplication::quit()` and `_exit(0)` on macOS, so a prompt raised there would never render. (I wired it there first and found this.)

## Privacy

Gated on **telemetry consent**: identity is `SentryReporter::anonymousInstallationId()`, which only exists when telemetry is on. Without it a response can't be joined to that install's activation/retention data — which is the only reason to ask. `Feedback/promptEnabled` (default ON) is a *separate* opt-out in Preferences, since a visible dismissible prompt isn't silent telemetry.

Import/export hooks pass `QFileInfo::suffix()` — **never the path**. No filenames, paths, or model content leave the machine.

## Transport

`submitFeedback` now accepts a bearer token **or** an `anonymousInstallationId`. `authorizedJsonRequest` omits the `Authorization` header entirely when there's no token — a bare `Bearer ` reads as a malformed credential rather than an unauthenticated request. `FeedbackDialog` can submit anonymously and no longer claims otherwise.

## Telemetry

`feedback.prompt_shown|dismissed|positive|negative|submitted` via `captureTelemetryEvent`, carrying workflow stage, session duration, file format, error category, and feedback category — all non-sensitive, all joinable to the existing install-level retention data.

## Tests

12 headless policy tests: telemetry gate, cooldown, dismissal backoff, one-per-session, the no-listener guard, first-import silence, failure prefill, and trigger-tag stability (those tags *are* telemetry queries — renaming one silently breaks saved dashboards).

⚠️ **They compile and link but have not been executed** — this session has no GL context and `UnitTests` requires one even for `--gtest_list_tests`. CI's Xvfb lane is their first real run.

Worth noting: the tests were initially **not compiled at all** — CMake's `*_test.cpp` glob was cached from before the file existed, so it was silently excluded while the build still reported success. Caught by grepping the binary for the test symbols (0 → 26 after reconfigure). Same class of trap as `QT_NO_WARNING_OUTPUT` silently deleting `qWarning` diagnostics.

## Deploy ordering

Migration 0028 must be applied before this ships, or the anonymous path 401s. It fails silently, so there's no hard ordering requirement — but no anonymous feedback is captured until then.

## Summary by CodeRabbit

* **New Features**
  * Added contextual feedback prompts after relevant import, export, and session activity.
  * Added a preference to enable or disable occasional feedback requests.
  * Feedback can be submitted anonymously when telemetry is enabled.
  * Prompts offer quick ratings and optional detailed feedback with contextual information.

* **Privacy & Telemetry**
  * Added sanitized feedback lifecycle telemetry without collecting message text.
  * Feedback requests respect consent, cooldowns, and dismissal limits.

* **Documentation**
  * Documented feedback triggers, privacy controls, submission behavior, and safety considerations.


fernandotonon:
Thanks — all nine were real and are fixed in `aa735fca` (plus the link fix in `95276d90`). Two were serious enough to call out:

**@chatgpt-codex-connector — `canSubmit()` still required `m_signedIn` (P1).** You're right that this made the entire anonymous path unreachable: Send stayed disabled for exactly the signed-out users the prompt targets, so `onSendClicked()` returned before any of the new code could run. Validation now accepts a non-empty anonymous installation id as an identity, and the "sign in required" notice only appears when there is genuinely neither.

**@coderabbitai — free-form text to Sentry.** Correct, and I'd reasoned about the wrong threat: I checked that `sanitizedValue` redacts paths and filenames and concluded the mirror was safe, entirely missing prose PII like an email typed into a comment. Sentry now gets structured metadata plus `message_length`; the words stay in the cloud row, linked by `feedback_id`.

**Export success with no file** (both of you) — `exporter()` returns an empty path on a cancelled save dialog, declined texture flattening, or an empty selection, and `noteExport(true, …)` ran anyway. That marked `firstExportDone`, could claim "you just exported your first model", and suppressed the session-no-export trigger for a session that wrote nothing. Now gated on a non-empty path, reporting the real extension.

**`reportSubmitted` was dead code** — wired to `dialog.exec() == Accepted`, which `FeedbackDialog` returns only after a successful POST, carrying the category.

**Non-modal context race** — context is now frozen in `maybePrompt()`; prefill, lifecycle events and the silent rating all read that snapshot instead of the mutable `m_last*` fields.

**Test POSTing to the live API** — fixed at the source rather than in the test: `postSilentRating` now refuses unless a prompt was actually shown this session, so no stray `reportPositive()` call can create synthetic rows in production. The test no longer calls it either.

Also fixed: scene-export failures pass the real format, and the CLAUDE.md claim that "every outcome persists to the cloud" was wrong — dismissals are Sentry-only by design, and a negative answer only creates a row if the detailed dialog is completed. Documented per-outcome.

fernandotonon:
Good catch — fixed in `0ee066dc`. You were right that gating `noteExport` but not the adjacent workflow event left cancelled exports counted as completions, which is exactly the metric this feature exists to measure.

Implementing it surfaced two further problems worth recording:

1. **`file.export.cancelled` was not in `isKnownTelemetryEvent`'s allow-list**, so `captureTelemetryEvent` would have dropped it *silently* — the new event would simply never have appeared in Sentry.
2. **`captureFileWorkflowEvent` maps `success=false` → `error` level**, so a user closing a save dialog would have been recorded as an error. Cancellation now keeps `success=false` (so the funnel doesn't count it as a completion) but logs at `info`; genuine failures stay `error`.

Added two regression tests. The cancelled one fails with a message pointing at the allow-list, since a missing entry there is otherwise invisible — the event just vanishes.

